PR #1, opened 2021-01-19
Replace CountMinSketchU8 with FrequencySketch 

tatsuya6502:
Replace `CountMinSketchU8` from count-min-sketch crate with `FrequencySketch`, which is basically a port from Caffeine.

- Bumped the version to 0.1.0-alpha.2.
- Ported frequency sketch from Caffeine, which supports aging.
- Removed count-min-sketch crate from the dependencies.
- Now the cache stores hash values of the keys in the deques, so that `admit()` function no longer has to recalculate hashes for victim entries.
- Now the hash calculations are done by client threads, not the draining thread.


tatsuya6502 (commented):
They look good to me.

PR #4, opened 2021-01-25
Reorganize modules

tatsuya6502:
- Move cache, segment (cache) and builder module to sync module.
- Move some types and traits out from cache module to dedicated modules.
- Update quanta to 0.7.1, which has a fix for compile errors under non-x86_64 platforms (such as arm64).


tatsuya6502 (commented):
LGTM

Issue #10, opened 2021-02-26
Ability to invalidate the entire cache?


jamesbirtles:
A use case has come up where I'd like to invalidate the entire cache (or perhaps invalidate based on a predicate on the values). I'm not seeing any way to do this at the moment. I am using the future based cache.

tatsuya6502:
Hi. I started to implement `invalidate_all` method, which invalidate the entire cache. (#11)

Once I am done with it, I will start to work on `invalidate_entries_if` method, which takes a predicate and invalidate entries that meet the predicate.

I will let you know when they are ready.

UnwrittenFun:
Awesome! 🎉

tatsuya6502:
@UnwrittenFun &mdash; I just published [Moka 0.3.0](https://crates.io/crates/moka/0.3.0) to crates.io. It has `invalidate_all` method and I hope you will like it. ([API doc](https://docs.rs/moka/0.3.0/moka/future/struct.Cache.html#method.invalidate_all))

I will keep this issue open until I finish implementing `invalidate_entries_if` method.

tatsuya6502:
> I will keep this issue open until I finish implementing `invalidate_entries_if` method.

Finally, I finished implementing this method (PR #12). It takes a predicate and invalidate entries that satisfy the predicate. (`key: &K` and `value: &V` are passed to the predicate)  The invalidation is processed by a background thread and client reads/writes will not be blocked while running the invalidation.

I am running some load tests now. If they go well, I will merge the PR and release Moka v0.4.0.

PR #11, opened 2021-02-27
Add the `invalidate_all` method to all caches

tatsuya6502:
This pull request adds the `invalidate_all` method to all `sync`, `future` and `unsync` caches. The `invalidate_all` method was requested in #10.

- Add `invalidate_all` method.
- Update internal `handle_upsert` and `evict` methods to prevent problems including segfault caused by race conditions between the sync/eviction thread and client writes (especially `invalidate`).
- Bump the version to 0.3.0.
- Upgrade a dependency: once_cell 1.5 -> 1.7.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,7 +10,7 @@
         "Moka",
         "Ristretto",
         "Tatsuya",
-        "unsync",
+        "Upsert",
         "actix",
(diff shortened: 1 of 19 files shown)


tatsuya6502 (commented):
They look good to me.

PR #13, opened 2021-04-02
Raise MSRV for "future" feature to 1.46.0

tatsuya6502:
Moka's minimum supported Rust version (MSRV) is currently 1.45.2. However, it does not compile anymore on 1.45.2 because the latest version of socket2 crate (0.4.0) requires Rust 1.46.0 or newer. 

Moka indirectly depends on socket2 via async-io or async-std crates:

1. Moka depends on async-io only when "future" feature is enabled.
    - Therefore, we will raise the MSRV for future feature to 1.46.0.
2. Moka depends on async-std only for testing (a dev dependency).
    - We want to keep the MSRV as 1.45.2 when no feature is enabled.

This PR does the followings:

- For 1 (MSRV 1.46.0 for future feature):
    - Update a CI (GitHub Action) job for Rust 1.45.2 to skip a step with `cargo test --release --features future`.
    - Add a CI job for Rust 1.46.0 to run the same step.
- For 2 (MSRV 1.45.2 for no feature enabled), updates `Cargo.toml` to disable default features on async-std.
- Update the README and Rustdoc to explain the latest MSRVs.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,7 +21,8 @@ jobs:
         rust:
           - stable
           - beta
-          - 1.45.2  # MSRV
+          - 1.46.0  # MSRV (future)
+          - 1.45.2  # MSRV (no features)
 
     steps:
       - name: Checkout Moka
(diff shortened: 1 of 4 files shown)


tatsuya6502 (commented):
All CI checks have passed. Merging.

PR #16, opened 2021-06-03
Stop skeptic from having to be compiled by all downstream users

paolobarbolini:
This stops skeptic from having to be included in the build stage of all downstream users of this crate (which I'm currently getting because of [libunftp](https://crates.io/crates/libunftp) depending on this).

skeptic is a [pretty heavy crate](https://crates.io/crates/skeptic/0.13.6/dependencies), depending on other crates like `cargo_metadata`, `error_chain`, `pulldown_cmark` ecc. which in turn depend on a bunch of other stuff. This doesn't make sense to me, as tests for an external crates shouldn't be run by downstream users.

This PR makes it so skeptic tests are only run when compiling with the `RUSTFLAGS="--cfg skeptic"` flag

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,6 +15,9 @@ on:
     # Run against the last commit on the default branch on Friday at 8pm (UTC?)
     - cron:  '0 20 * * 5'
 
+env:
+  RUSTFLAGS: '--cfg skeptic'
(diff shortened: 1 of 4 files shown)


tatsuya6502 (approved):
Thank you for submitting the PR. LGTM.

PR #19, opened 2021-06-13
Add Send + Sync + 'static bounds to K, V and S of sync and future Caches

tatsuya6502:
This PR adds `Send`, `Sync` and `'static` bounds to the type parameters `K` (key), `V` (value) and `S` (hasher state) of the following cache implementations:

- `sync::Cache`
- `sync::SegmentedCache`
- `future::Cache`

This will prevent the following usage from drawing undefined behavior:

- Single thread application using an above thread-safe cache.
- Store non-`Send` or non-`Sync` type (such as `std::rc::Rc`) as key (`K`) and/or (`Value`).
- Or, store non static reference type as key and/or value.

Until now, storing such key or value types was possible in _single thread_ application (including single-threaded async runtime such as Actix-rt) because above caches did not ask these bounds explicitly. 

However doing so is incorrect as it can draw undefined behavior. This is because these caches use multiple threads internally for updating internal data structures and evicting cached values.

```rust
use std::{rc::Rc, time::Duration};

// This code will compile, but cat hit an undefined behavior (UB) as breaking
// thread safety guarantee of std::rc::Rc.

// Create a sync::Cache with TTL 1 second.
let cache = moka::sync::CacheBuilder::new(100)
    .time_to_live(Duration::from_secs(1))
    .build();

// Create a Rc, which is !Send and !Sync. (Rc's strong reference count should be 1)
let v = Rc::new("Hello".to_string());

// Insert a clone of RC to the cache. (The strong count should be 2 now)
cache.insert(0, Rc::clone(&v));

// Sleep for 5 seconds.
std::thread::sleep(Duration::from_secs(5));

// While sleeping, the value `v` should have been expired so sync::Cache's
// eviction thread will drop it. Dropping will make the strong count to be
// decremented by 1 but this thread may not see the decrement as Rc is
// neither Send nor Sync.

println!("{}", Rc::strong_count(&v));  // 1 or 2? (*1)
```

Adding `Send`, `Sync` and `'static` bounds should prevent the problem by making such code not to compile by type errors.

Note that multi-thread APIs such as `std::thread::spawn` or multi-threaded async runtime (such as the ones provided by Tokio and async-std) will ask those bounds explicitly. So there should have been no problem with them.

----

*1: I wrote the code for illustrating purpose. If you actually run the code, it will always prints `2`. However, this does not mean you hit the UB, but it is actually an expected behavior. `sync::Cache` and `future::Cache` employ concurrent hash table `cht::SegmentedHashMap` as the main key value storage. cht will not drop a removed value immediately by the spec to provide high concurrency and thread safety. See this issue for more details: https://github.com/Gregory-Meyer/cht/issues/23

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Moka &mdash; Change Log
 
+## Version 0.4.0
+
+### Fixed
+
+- **Breaking change**: Now `sync::{Cache, SegmentedCache}` and `future::Cache`
+  require `Send`, `Sync` and `'static` bounds for the generic parameters `K` (key),
+  `V` (value) and `S` (hasher state). This is necessary to prevent potential
+  undefined behaviors in applications using single-threaded async runtime such as
+  Actix-rt. ([#19][gh-pull-0019])
+
+
 ## Version 0.3.1
 
 ### Changed
@@ -49,6 +60,7 @@
 
 [caffeine-git]: https://github.com/ben-manes/caffeine
 
+[gh-pull-0019]: https://github.com/moka-rs/moka/pull/19/
 [gh-pull-0016]: https://github.com/moka-rs/moka/pull/16/
 [gh-pull-0011]: https://github.com/moka-rs/moka/pull/11/
 [gh-pull-0009]: https://github.com/moka-rs/moka/pull/9/
--- a/src/future/builder.rs
+++ b/src/future/builder.rs
@@ -47,8 +47,8 @@ pub struct CacheBuilder<C> {
 
 impl<K, V> CacheBuilder<Cache<K, V, RandomState>>
 where
-    K: Eq + Hash,
-    V: Clone,
+    K: Eq + Hash + Send + Sync + 'static,
+    V: Clone + Send + Sync + 'static,
 {
     /// Construct a new `CacheBuilder` that will be used to build a `Cache` holding
     /// up to `max_capacity` entries.
@@ -78,7 +78,7 @@ where
     /// Builds a `Cache<K, V, S>`, with the given `hasher`.
     pub fn build_with_hasher<S>(self, hasher: S) -> Cache<K, V, S>
     where
-        S: BuildHasher + Clone,
+        S: BuildHasher + Clone + Send + Sync + 'static,
     {
         Cache::with_everything(
             self.max_capacity,
--- a/src/future/cache.rs
+++ b/src/future/cache.rs
@@ -209,8 +209,8 @@ where
 
 impl<K, V> Cache<K, V, RandomState>
 where
-    K: Hash + Eq,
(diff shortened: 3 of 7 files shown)


tatsuya6502:
> Adding `Send`, `Sync` and `'static` bounds should prevent the problem by making such code not to compile by type errors.

Verified that the above code no longer compiles.

```console
error[E0277]: `Rc<String>` cannot be sent between threads safely
 --> examples/rc-ex.rs:8:13
  |
8 | let cache = moka::sync::CacheBuilder::new(100)
  |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `Rc<String>` cannot be sent between threads safely
  |
  = help: the trait `Send` is not implemented for `Rc<String>`
  = note: required by `moka::sync::CacheBuilder::<moka::sync::Cache<K, V>>::new`

error[E0277]: `Rc<String>` cannot be shared between threads safely
 --> examples/rc-ex.rs:8:13
  |
8 | let cache = moka::sync::CacheBuilder::new(100)
  |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `Rc<String>` cannot be shared between threads safely
  |
  = help: the trait `Sync` is not implemented for `Rc<String>`
  = note: required by `moka::sync::CacheBuilder::<moka::sync::Cache<K, V>>::new`

error[E0277]: `Rc<String>` cannot be sent between threads safely
  --> examples/rc-ex.rs:10:6
   |
10 |     .build();
   |      ^^^^^ `Rc<String>` cannot be sent between threads safely
   |
   = help: the trait `Send` is not implemented for `Rc<String>`

error[E0277]: `Rc<String>` cannot be shared between threads safely
  --> examples/rc-ex.rs:10:6
   |
10 |     .build();
   |      ^^^^^ `Rc<String>` cannot be shared between threads safely
   |
   = help: the trait `Sync` is not implemented for `Rc<String>`

error[E0277]: `Rc<String>` cannot be sent between threads safely
  --> examples/rc-ex.rs:16:17
   |
16 | cache.insert(0, Rc::clone(&v));
   |                 ^^^^^^^^^^^^^ `Rc<String>` cannot be sent between threads safely
   |
   = help: the trait `Send` is not implemented for `Rc<String>`

error[E0277]: `Rc<String>` cannot be shared between threads safely
  --> examples/rc-ex.rs:16:17
   |
16 | cache.insert(0, Rc::clone(&v));
   |                 ^^^^^^^^^^^^^ `Rc<String>` cannot be shared between threads safely
   |
   = help: the trait `Sync` is not implemented for `Rc<String>`

error: aborting due to 6 previous errors

For more information about this error, try `rustc --explain E0277`.
error: could not compile `moka`

To learn more, run the command again with --verbose.


```

tatsuya6502 (commented):
Merging.

PR #20, opened 2021-06-26
Add `get_or_insert_with` method

### LINKED ISSUE #15: Feature Request: Get or Insert Future
The ability to get-or-insert atomically would be nice for preventing multiple requests for the same resource from being created to populate the cache. Otherwise, there's a race condition between get and insert where another task could attempt to get as well, and also see an empty record to populate.

tatsuya6502:
This PR adds the `get_or_insert_with` and `get_or_try_insert_with` methods to `sync` and `future` caches in Moka. These methods ensure the key exists in the cache by inserting the return value from the given closure/future if the key does not exist. They also guarantee the closure/future is evaluated only once even if they are called from multiple threads/async tasks for the same key at the same time.

This feature was requested by #15.

## Added

- (API) Add `get_or_insert_with` and `get_or_try_insert_with` methods to the following cache implementations in Moka:
    - `sync::Cache`
    - `sync::SegmentedCache`
    - `future::Cache`
- Add internal `sync::value_initializer` and `future::value_initializer` modules, containing `cht::HashMap` for waiters.

----
Fixes #15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Moka &mdash; Change Log
 
+## Version 0.5.0 (Not released yet)
+
+### Added
+
+- Add `get_or_insert_with` and `get_or_try_insert_with` methods to `sync` and
+  `future` caches. ([#20][gh-pull-0020])
+
+
 ## Version 0.4.0
 
 ### Fixed
@@ -14,6 +22,7 @@
 
 - Add `invalidate_entries_if` method to `sync`, `future` and `unsync` caches.
   ([#12][gh-pull-0012])
+
 
 ## Version 0.3.1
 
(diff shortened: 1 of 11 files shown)


tatsuya6502 (commented):
Merging.

ben-manes:
I don't know rust conventions, but Java's names in `Map` are quite nice.
- computeIfAbsent(key, key -> value)
- computeIfPresent(key, (key, oldValue) -> value)
- compute(key, (key, oldValue) -> value
- merge(key, value, (key, oldValue) -> newValue)

These are performed atomically and if `null` is returned then the mapping is removed. The names are nice as succinct and easy to understand. The three compute methods are very useful for advanced cases.

Caffeine's corre apis uses `get` for automatic loading, `getIfPresent` without, and offers an `asMap()` view for richer functionality (like `compute` or entry iterators). Since we highly prefer users to delegate to the cache for loading, we take the desirable `get` name and penalize the longer name `getIfAbsent` to dissuade usage. A lot of random functionality, like conditional entry removal or policy inspection, is hidden away under nested objects (`asMap()`, `policy()`) to provide the features while keeping the core interface simple to reduce the conceptual weight for learning a new library.

I don't know if any of that makes sense idiomatically for Rust. If so, you might consider reviewing our APIs for ideas before yours are finalized.

PR #21, opened 2021-07-13
Enable hash-arg feature in moka-cht

tatsuya6502:
Utilize moka-cht's `hash-arg` feature to avoid calculating the same hash twice for a key at both Moka cache and cht sides.


tatsuya6502:
I will close this for now. I will tackle to this again once I am done with other high priority items.

PR #23, opened 2021-08-03
Change get_or_try_insert_with to return a concrete error type rather than a trait object

tatsuya6502:
Change `get_or_try_insert_with` of `sync::{Cache, SegmentedCache}` and `future::Cache` to return a concrete error type rather than a trait object.

**Before** (`sync::Cache`)

```rust
pub fn get_or_try_insert_with<F>(&self, key: K, init: F) ->
    Result<V, Arc<Box<dyn Error + Send + Sync + 'static>>>
where
    F: FnOnce() -> Result<V, Box<dyn Error + Send + Sync + 'static>>
```

**After**

```rust
pub fn get_or_try_insert_with<F, E>(&self, key: K, init: F) -> Result<V, Arc<E>>
where
    F: FnOnce() -> Result<V, E>,
    E: Error + Send + Sync + 'static,
```

## Behavior Change

There is a subtle change in the behavior. `get_or_try_insert_with` used to guarantee that the closure `F` to evaluate once even if multiple threads call it at the same time on the same cache key.

Now this is not always true:
- Only one of the closures must be evaluated if the closures in all concurrent calls on the same cache key return _the same_ concrete error type.
- One closure of each error type should be evaluated if the closures in concurrent calls on the same cache key does not return the same concrete error type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Moka &mdash; Change Log
 
+## Version 0.6.0 (Unreleased)
+
+### Changed
+
+- Change `get_or_try_insert_with` to return a concrete error type rather
+  than a trait object. ([#23][gh-pull-0023])
+
+
 ## Version 0.5.1
 
 ### Changed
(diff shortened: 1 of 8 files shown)


tatsuya6502 (commented):
Merging.

PR #28, opened 2021-08-28
Handle `usize` overflow on big cache capacity

### LINKED ISSUE #25: [BUG] `usize` overflow on big cache capacity
Hello,

First of all thank you for the very nice work !

There seems to be a multiply overflow that could be handled at the line referenced below.

https://github.com/moka-rs/moka/blob/76b06af686c5ecaa913abf1acb46e45fc4f0a95c/src/sync/base_cache.rs#L395

This could be done either by returning an error type, or by maxing the value of the cache capacity to `usize::max_value() / 32`

tatsuya6502:
This PR fixes the issue #25, `usize` multiply overflow on big cache capacity.

## Changes

- Removed the multiply operation (`max_capacity * 32`) when setting the capacity of the popularity estimator `FrequencySketch`.
    - The implementation of the estimator was replaced in an older version of Moka and the multiply operation was not really needed since then.
- Added code to ensure that the capacity of the popularity estimator always fits within a range of `128u32..=u32::MAX`.
- Changed the type of the internal fields of the estimator from `usize` to `u32` so that the estimator will work consistency on any platform with different pointer width.

## Verification

- Added unit tests to `sync::base_cache` and `unsync::cache` modules to ensure overflows should not happen and sizes of the population estimators are correct:
    - Before applying the fix, ran the tests with the debug profile and confirmed they panicked as expected:
      ```
      ---- sync::base_cache::tests::test_skt_capacity_will_not_overflow stdout ----
      thread 'sync::base_cache::tests::test_skt_capacity_will_not_overflow' panicked at
      'attempt to multiply with overflow', src/sync/base_cache.rs:395:39
      ```
    - After applying the fix, ran the tests again and verified they passed. **PASSED**
- Ran some mokabench workloads on Moka before/after the fix and verified that hit rates and run-time durations did not change. **PASSED**

----
Fixes #25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Moka &mdash; Change Log
 
+## Version 0.5.2
+
+### Fixed
+
+- Handle `usize` overflow on big cache capacity. ([#28][gh-pull-0028])
+
+
 ## Version 0.5.1
 
 ### Changed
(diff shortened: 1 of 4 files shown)


tatsuya6502 (commented):
Merging.

PR #32, opened 2021-09-04
Add `Send` and `'static` bounds to the init futures of `get_or_(try_)insert_with`

### LINKED ISSUE #31: `future::Cache::get_or_(try_)insert_with()` can lead UB as accepting `init` future that is not `Send` or `'static`.
In Moka v0.5.0 and v0.5.1, the following code fragments for `future::Cache` will compile. However Moka should reject them as they can lead undefined behavior (UB) in safe Rust.

**Example 1**

```rust
use moka::future::Cache;
use std::rc::Rc;

#[tokio::main]
async fn main() {
    let cache: Cache<_, String> = Cache::new(100);

    // Rc is !Send.
    let data = Rc::new("zero".to_string());
    let data1 = Rc::clone(&data);

    cache
        .get_or_insert_with(0, async move {
            // A data race may occur. 
            // The async block can be executed by a different thread
            // but Rc's internal reference counters are not thread safe.
            data1.to_string()
        })
        .await;

    println!("{:?}", data);
}
```

**Example 2**

```rust
use moka::future::Cache;

#[tokio::main]
async fn main() {
    let cache: Cache<_, String> = Cache::new(100);

    let data = "zero".to_string();
    {
        // Not 'static.
        let data_ref = &data;

        cache
            .get_or_insert_with(0, async {
                // This may become a dangling pointer.
                // The async block can be executed by a different thread so
                // the captured reference `data_ref` may outlive its value.
                data_ref.to_string()
            })
            .await;
    }

    println!("{:?}", data);
}
```

So `future::Cache`'s `get_or_insert_with()` and `get_or_try_insert_with()` need extra bounds `Send` and `'static` on the `init` future.

tatsuya6502:
- Add `Send` and `'static` bounds to the `init` futures.
- Add UI tests to ensure compile failures to occur for misuses of `get_or_(try_)insert_with`.

----
Fixes #31


tatsuya6502 (commented):
Merging.

Issue #34, opened 2021-09-05
Segmentation faults in moka-cht under heavy workloads on a many-core machine


tatsuya6502:
I have seen segmentation faults a few times when I am running [mokabench](https://github.com/moka-rs/mokabench) on Moka v0.5.1. It seems it is randomly happening while `get_or_insert_with` method is heavily called concurrently from many threads.

```console
+ ./target/release/mokabench --enable-invalidate-entries-if --enable-insert-once
Cache, Max Capacity, Clients, Inserts, Reads, Hit Rate, Duration Secs
Moka Unsync Cache, 100000, -, 14696832, 31104534, 52.750, 8.575
Moka Cache, 100000, 16, 15550290, 31954711, 51.336, 17.365
Moka Cache, 100000, 24, 15543954, 31948375, 51.347, 17.743
Moka Cache, 100000, 32, 15527876, 31932297, 51.373, 17.877
./run-tests.sh: line 36: 21740 Segmentation fault      (core dumped) ./target/release/mokabench --enable-invalidate-entries-if --enable-insert-once
```

I am using Amazon EC2 for running mokabench. After spending few days, I found it is related to the version of [crossbeam-epoch](https://crates.io/crates/crossbeam-epoch) and number of CPU cores. 

| Segfaults? | Moka    | cht/moka-cht    | crossbeam-epoch | EC2 Instance Type | Arch   | vCPUs | OS             |
|:----------:|:-------:|:---------------:|:---------------:|:-----------------:|:------:|:-----:|----------------|
|     Yes    | v0.5.1  | moka-cht v0.5.0 |      v0.9.5     |    c5.9xlarge     | x86_64 |  36   | Amazon Linux 2 |
|     No     | v0.5.1  |   cht v0.4.1    |      **v0.8.2**     |    c5.9xlarge     | x86_64 |  36   | Amazon Linux 2 |
|     No     | v0.5.1  | moka-cht v0.5.0 |      v0.9.5     |    c5.4xlarge     | x86_64 |  **16**   | Amazon Linux 2 |

crossbeam-epoch is used by moka-cht, the concurrent hash table use by Moka.

I examined stack traces from core dumps and found there are two patterns. I have not identified the root cause yet. Perhaps a `crossbeam_epoch::Owned<T>`, which is very similar to `Box<T>`, stored in moka-cht became a dangling pointer by some reason?

<details>
[...]

PR #39, opened 2021-09-09
Support Linux on MIPS and ARMv5TE

### LINKED ISSUE #38: Support Linux on MIPS and ARMv5TE
Support the following 32-bit Linux platforms if possible:
- `mipsel-unknown-linux-musl` (for https://github.com/messense/aliyundrive-webdav/issues/22)
- `armv5te-unknown-linux-musleabi` (for https://github.com/messense/aliyundrive-webdav/issues/23)

Currently Moka will not compile by the following reasons:
- Moka uses `std::atomic::AtomicU64` mainly as the expiration clock for entries, but `AtomicU64` is not available on mipsel.
    - _"PowerPC and MIPS platforms with 32-bit pointers do not have `AtomicU64` or `AtomicI64` types"_ ([doc](https://doc.rust-lang.org/stable/std/sync/atomic/index.html#portability))
    - It seems `AtomicU64` is not available on ARMv5TE too.
- quanta will not compile too because crossbeam-utils does not provide `atomic::AtomicCell::fetch_add` method on these platforms: (https://github.com/metrics-rs/quanta/issues/54)

Probably we will replace all usages of `AtomicU64` with `RwLock<Option<std::time::Instant>>` on these platforms?

tatsuya6502:
This pull request is for supporting Linux on 32-bit MIPS and ARMv5TE. These platforms does not provide `std::sync::atomic::AtomicU64` so currently both Moka and quanta will not compile.

~~After merging this PR, Moka will compile on 32-bit MIPS. For ARMv5TE, you need to add `default-features = false` to "moka" dependency in Cargo.toml and then Moka will compile.~~

After this PR is merged, you can resolve the errors by disabling the default features by adding `default-features = false` to the dependency declaration in Cargo.toml. This will remove quanta from the dependencies and enable a fall-back implementation in Moka.

## Changes

- Add a cargo feature ~~`quanta`~~ `atomic64` and make it a default feature.
- Add a module `common::time` and conditionally choose the source file from "time_quanta.rs" or "time_compat.rs".
    - time_quanta.rs file has type aliases to `quanta::{Instant, Clock, Mock}`. It also has `AtomicInstant` holding an `AtomicU64`.
    - time_compat.rs file has `Instant` based on `std::time::Instant` and quanta-like `Clock` and `Mock`. It also has `AtomicInstant` holding a `RwLock<Option<Instant>>`.
- Update Cache, etc. modules to use `common::time` instead of `quanta`.
- Add a step to GitHub Actions to test future feature without quanta/`AtomicU64`.

---
Fixes #38.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -68,7 +68,7 @@ jobs:
         env:
           RUSTFLAGS: '--cfg skeptic'
 
-      - name: Run tests (future, without quanta)
+      - name: Run tests (future, without atomic64)
         uses: actions-rs/cargo@v1
         if: ${{ matrix.rust != '1.45.2' }}
         with:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -68,6 +68,15 @@ jobs:
         env:
           RUSTFLAGS: '--cfg skeptic'
 
+      - name: Run tests (future, without quanta)
+        uses: actions-rs/cargo@v1
+        if: ${{ matrix.rust != '1.45.2' }}
+        with:
+          command: test
+          args: --release --no-default-features --features future
+        env:
+          RUSTFLAGS: '--cfg skeptic'
+
       - name: Run UI tests (future, trybuild)
         uses: actions-rs/cargo@v1
         if: ${{ matrix.rust == 'stable' }}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,11 +1,13 @@
 {
     "rust-analyzer.cargo.features": ["future"],
     "cSpell.words": [
+        "aarch",
         "actix",
         "ahash",
         "benmanes",
         "CLFU",
         "clippy",
+        "compat",
         "cpus",
         "deqs",
         "Deque",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Added
 
 - Add support for some 32-bit platforms where `std::sync::atomic::AtomicU64` is not
-  provided. (e.g. `armv5te-unknown-linux-musleabi`, `mips-unknown-linux-musl`)
+  provided. (e.g. `armv5te-unknown-linux-musleabi` or `mips-unknown-linux-musl`)
     - On these platforms, you will need to disable the default features of Moka.
       See [the relevant section][resolving-error-on-32bit] of the README.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Moka &mdash; Change Log
 
+## Version 0.5.3
+
+### Added
+
+- Add support for some 32-bit platforms where `std::sync::atomic::AtomicU64` is not
+  provided. (e.g. `armv5te-unknown-linux-musleabi`, `mips-unknown-linux-musl`)
+    - On these platforms, you will need to disable the default features of Moka.
+      See [the relevant section][resolving-error-on-32bit] of the README.
+
+
 ## Version 0.5.2
 
 ### Fixed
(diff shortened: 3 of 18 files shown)


messense:
Note that I've also opened https://github.com/metrics-rs/quanta/pull/55 to fix quanta on these platforms, so maybe after that lands moka can still use quanta but not `AtomicU64`.

tatsuya6502:
Thanks. Good point.

Hmm. I am having hard time to come up with a good _short_ name for the feature. Moka uses `AtomicU64` as a lightweight atomic type for timestamps. It is for space efficiency and also high performance.

So... perhaps, `atomic-time`, ... or just `atomic-u64`/`atomic64`?

I think `atomic64` might be OK. It may be too straightforward, but works. It will be easy to understand that you need to disable `atomic64` feature if platform does not support 64-bit atomic types.

messense (approved):
Thank you.

messense:
I've just tested this on a mipsel router, it compiles and runs, but hits https://github.com/seanmonstar/num_cpus/issues/69 , I think this line should be changed use 1 if `num_cpus::get()` returns 0

https://github.com/moka-rs/moka/blob/988816ad77f6192ba6d7d96a3bfd534cf0206d80/src/common/thread_pool.rs#L56

tatsuya6502:
Thanks for testing on the real hardware. I did not realize it could return 0. Changed to use 1 when it returns 0.

tatsuya6502:
Verified the followings:

1. Ran `cargo build` for the following targets using Docker images `messense/rust-musl-cross` and verified they succeeded:
    - `armv5te-unknown-linux-musleabi`
    - `mips-unknown-linux-musl` 
    - `mipsel-unknown-linux-musl`
    -  commit: https://github.com/moka-rs/moka/commit/d91c5ce4bbd6d403e1fe37fa9ed1831225b18434
2. Ran regular prerelease load tests using mokabench on `x86_64-unknown-linux-gnu`. Verified that the test runs with and without the default features found no new issue.
    - commit: https://github.com/moka-rs/moka/commit/4407e87ecfde260ed39ddded5ee9b517c0757fa5
    - NOTE: This known issue is still occurring: #34

**Test 1**

```console
% git remote -v
origin	git@github.com:moka-rs/moka.git (fetch)
origin	git@github.com:moka-rs/moka.git (push)

$ git log -1
commit d91c5ce4bbd6d403e1fe37fa9ed1831225b18434 (HEAD -> mips-armv5, origin/mips-armv5)
Author: Tatsuya Kawano
Date:   Fri Sep 10 20:19:23 2021 +0800

    Bump the version to v0.5.3


$ sudo rm -rf release

$ docker run -it --rm -v $(pwd):/home/rust/src \
  --mount source=cargo-registry,target=/root/.cargo/registry \
  messense/rust-musl-cross:armv5te-musleabi \
  cargo build --no-default-features --features future --target armv5te-unknown-linux-musleabi
   ...
   Compiling moka v0.5.3 (/home/rust/src)
   ...
    Finished dev [unoptimized + debuginfo] target(s) in 32.67s

$ echo $?
0

$ docker run -it --rm -v $(pwd):/home/rust/src \
  --mount source=cargo-registry,target=/root/.cargo/registry \
  messense/rust-musl-cross:mips-musl \
  sh -c "RUSTFLAGS='-C target-feature=+crt-static -C link-arg=-s' \
    cargo build --no-default-features --features future --target mips-unknown-linux-musl"
   ...
   Compiling moka v0.5.3 (/home/rust/src)
   ...
    Finished dev [unoptimized + debuginfo] target(s) in 8.96s
$ echo $?
0

$ docker run -it --rm -v $(pwd):/home/rust/src \
  --mount source=cargo-registry,target=/root/.cargo/registry \
  messense/rust-musl-cross:mipsel-musl \
  sh -c "RUSTFLAGS='-C target-feature=+crt-static -C link-arg=-s' \
    cargo build --no-default-features --features future --target mipsel-unknown-linux-musl"
   ...
   Compiling moka v0.5.3 (/home/rust/src)
   ...
    Finished dev [unoptimized + debuginfo] target(s) in 8.82s
$ echo $?
0
```

messense:
Test again on mipsel router and I'm happy to report that it works™.

tatsuya6502:
Thank you for testing. I am merging this and will publish v0.5.3 to crate.io soon.

PR #40, opened 2021-09-11
CI: Add cross-compile tests for some Linux targets

tatsuya6502:
Add a CI (GitHub Actions) using [rust-embedded/cross](https://github.com/rust-embedded/cross). It does the followings:

1. Cross-compile Moka to a specified Linux target.
2. Run `cross test`(similar to `cargo test` but without integration tests) using QEMU user space emulation.

Currently, the following targets are tested:

- `aarch64-unknown-linux-musl`
- `i686-unknown-linux-musl`
- `armv7-unknown-linux-musleabihf`
- `armv5te-unknown-linux-musleabi`
- `mips-unknown-linux-musl`
- `mipsel-unknown-linux-musl`

There is one concern though. [The README of cross](https://github.com/rust-embedded/cross#supported-targets) says: 

> `cross test` runs units tests sequentially because QEMU gets upset when you spawn multiple threads. This means that, if one of your unit tests spawns threads, then it's more likely to fail or, worst, never terminate.

Moka _does_ spawn threads but it seems QEMU is working fine. So it is OK for now. If such a problem happens in the future, we will have to disable the `cross test` step.

### DIFF
--- a/.github/workflows/LinuxCrossCompileTest.yml
+++ b/.github/workflows/LinuxCrossCompileTest.yml
@@ -1,0 +1,73 @@
+name: Linux cross compile tests
+
+on:
+  push:
+    paths-ignore:
+    - '.devcontainer/**'
+    - '.gitpod.yml'
+    - '.vscode/**'
+    - 'tests/**'
+  pull_request:
+    paths-ignore:
+    - '.devcontainer/**'
+    - '.gitpod.yml'
+    - '.vscode/**'
+    - 'tests/**'
+  schedule:
+    # Run against the last commit on the default branch on Friday at 9pm (UTC?)
+    - cron:  '0 21 * * 5'
+
+jobs:
+  linux-cross:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        platform:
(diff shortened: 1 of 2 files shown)


tatsuya6502 (commented):
Merging.

tatsuya6502:
> There is one concern though. [The README of cross](https://github.com/rust-embedded/cross#supported-targets) says: 
> 
>> `cross test` runs units tests sequentially because QEMU gets upset when you spawn multiple threads. This means that, if one of your unit tests spawns threads, then it's more likely to fail or, worst, never terminate.
> 
> Moka _does_ spawn threads but it seems QEMU is working fine. So it is OK for now. If such a problem happens in the future, we will have to disable the `cross test` step.

I realized the README was written in December 2016. So the QEMU's user-space emulatiton may have been improved since then. This could explain why `cross test` is working for me.

- https://github.com/rust-embedded/cross/commit/531a8b8880178df2480a240ed4261263f78c51c0#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R68-R70

PR #46, opened 2021-09-18
Handle panic in the `init` closure/future of `get_or_insert_with` and `get_or_try_insert_with`

### LINKED ISSUE #43: Panic causes "thread panicked at 'internal error: entered unreachable code'"
Hi, thank you for this very useful library! However, I've found a way to enter a supposedly unreachable state.

Here's the relevant part of the backtrace:

```
thread 'main' panicked at 'internal error: entered unreachable code', /home/user/.cargo/git/checkouts/moka-6ea430727379b61e/19b6925/src/future/value_initializer.rs:101:29
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:515:5
   1: core::panicking::panic_fmt
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/panicking.rs:50:5
   3: moka::future::value_initializer::ValueInitializer<K,V,S>::try_init_or_read::{{closure}}
             at /home/user/.cargo/git/checkouts/moka-6ea430727379b61e/19b6925/src/future/value_initializer.rs:101:29
   4: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/future/mod.rs:80:19
   5: moka::future::cache::Cache<K,V,S>::get_or_try_insert_with_hash_and_fun::{{closure}}
             at /home/user/.cargo/git/checkouts/moka-6ea430727379b61e/19b6925/src/future/cache.rs:640:15
   6: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/future/mod.rs:80:19
   7: moka::future::cache::Cache<K,V,S>::get_or_try_insert_with::{{closure}}
             at /home/user/.cargo/git/checkouts/moka-6ea430727379b61e/19b6925/src/future/cache.rs:445:9
```

This issue seems to occur when two threads attempt to fetch the same key and the thread on which the value initialization was started panics. I've built a simple program to demonstrate this issue:

```rust
use tokio;
use std::sync::Arc;

#[tokio::main]
async fn main() {
    let cache: Arc<moka::future::Cache<i32, i32>> = Arc::new(moka::future::Cache::new(16));
    let semaphore = Arc::new(tokio::sync::Semaphore::new(0));
    {
        let cache_ref = cache.clone();
        let semaphore_ref = semaphore.clone();
        tokio::task::spawn(async move {
            cache_ref.get_or_try_insert_with::<_, ()>(1, async move {
                semaphore_ref.add_permits(1);
                tokio::time::sleep(tokio::time::Duration::from_millis(1000));
                panic!("Panic during get_or_try_insert_with");
                Ok(10)
            }).await;
        });
    }
    semaphore.acquire().await;
    cache.get_or_try_insert_with::<_, ()>(1, async move {
        println!("Async block in second get_or_try_insert_with called");
        Ok(5)
    }).await;
}
```

I think the unreachable code is being hit because the write lock obtained at [line 71 of value_initializer.rs](https://github.com/moka-rs/moka/blob/19b6925d9e854374f97adecb074ddc192c47ee6b/src/sync/value_initializer.rs#L71) gets dropped without any value being written ([here](https://github.com/moka-rs/moka/blob/19b6925d9e854374f97adecb074ddc192c47ee6b/src/sync/value_initializer.rs#L78) or [here](https://github.com/moka-rs/moka/blob/19b6925d9e854374f97adecb074ddc192c47ee6b/src/sync/value_initializer.rs#L83)).

I see two possible options for handling this issue:
1. `unreachable()` can be replaced with a panic with a more helpful message.
2. Perhaps the existing waiter should be removed and a new waiter should be created and started using the future passed in the second call?

tatsuya6502:
This PR addresses the issue reported by #43.

## Changes

When the user-provided `init` future/closure for `get_or_insert_with` or `get_or_try_insert_with` method has panicked, it will catch the panic, remove the waiter for the key, and resume the panic. It also restarts other in-progress (actually blocked) method calls on the same key.

- Add `futures` crate for Moka's `future` feature.
- Update `future::Cache` to handle panic in the `init` future of `get_or_insert_with` and `get_or_try_insert_with`.
- Update `sync::Cache` to handle panic in the `init` future of `get_or_insert_with` and `get_or_try_insert_with`.
- Add unit tests.

----
Fixes #43

### DIFF
--- a/src/future/value_initializer.rs
+++ b/src/future/value_initializer.rs
@@ -82,7 +82,7 @@ where
 
     /// # Panics
     /// Panics if the `init` future has been panicked.
-    async fn do_try_init<'a, F, FO, C, E>(
+    async fn do_try_init<'a, F, O, C, E>(
         &self,
         key: &'a Arc<K>,
         type_id: TypeId,
@@ -90,8 +90,8 @@ where
         mut post_init: C,
     ) -> InitResult<V, E>
     where
-        F: Future<Output = FO>,
-        C: FnMut(&'a Arc<K>, FO, &mut WaiterValue<V>) -> InitResult<V, E>,
+        F: Future<Output = O>,
+        C: FnMut(&'a Arc<K>, O, &mut WaiterValue<V>) -> InitResult<V, E>,
         E: Send + Sync + 'static,
     {
         use futures::future::FutureExt;
@@ -107,15 +107,15 @@ where
                     // Our waiter was inserted. Let's resolve the init future.
                     // Catching panic is safe here as we do not try to resolve the future again.
                     match AssertUnwindSafe(init).catch_unwind().await {
-                        // resolved.
+                        // Resolved.
                         Ok(value) => return post_init(key, value, &mut lock),
-                        // panicked.
+                        // Panicked.
                         Err(payload) => {
                             *lock = None;
                             // Remove the waiter so that others can retry.
                             self.remove_waiter(key, type_id);
                             resume_unwind(payload);
-                        }
+                        } // The lock will be unlock here.
                     }
                 }
                 Some(res) => {
--- a/src/future/value_initializer.rs
+++ b/src/future/value_initializer.rs
@@ -98,6 +98,9 @@ where
         use std::panic::{resume_unwind, AssertUnwindSafe};
         use InitResult::*;
 
+        const MAX_RETRIES: usize = 200;
+        let mut retries = 0;
+
         loop {
             let waiter = Arc::new(RwLock::new(None));
             let mut lock = waiter.write().await;
@@ -115,7 +118,7 @@ where
                             // Remove the waiter so that others can retry.
                             self.remove_waiter(key, type_id);
                             resume_unwind(payload);
-                        } // The lock will be unlock here.
+                        } // The lock will be unlocked here.
                     }
                 }
                 Some(res) => {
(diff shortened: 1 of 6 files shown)


src/future/value_initializer.rs
```diff
@@ -35,70 +36,99 @@ where
...
+                Some(res) => {
+                    // Somebody else's waiter already exists. Drop our write lock and wait
+                    // for a read lock to become available.
+                    std::mem::drop(lock);
+                    match &*res.read().await {
+                        Some(Ok(value)) => return ReadExisting(value.clone()),
+                        Some(Err(e)) => return InitErr(Arc::clone(e).downcast().unwrap()),
+                        // None means somebody else's init future has been panicked.
+                        // Retry from the beginning.
+                        None => continue,
```
tatsuya6502, line 130: Just to be safer, we should avoid to create a busy, infinite loop here. Maybe we can do some of the followings:

- Add a retry limit. If the number of retries has reached to the limit, do panic with some helpful message.
- If it is still getting `None` after some retries, it probably means other `init` futures are also panicking. But is there any other reason? e.g. removing the waiter is failing? If so, is there anything we can do here?
    - Adding a short sleep between retries to avoid busy loop?
    - Try to remove the waiter here?

I think we just want to do the first one for now, and see what happens.

tatsuya6502: > Add a retry limit. If the number of retries has reached to the limit, do panic with some helpful message.

Added the retry limit: https://github.com/moka-rs/moka/pull/46/commits/af44363001962d92b425f6a139d2dc82f0d242ce

tatsuya6502 (commented):
Merging.

nickguletskii:
Thank you for the quick fix!

PR #56, opened 2021-12-16
protect overflow when computing expiration

barkanido:
This change protects against overflow using 2 measures:
1. cache builders now won't build caches with time_tolive or
time_to_idle higher than 1000 years.
2. for future  computations, the internal Instant type will now only
allow checked additions to force handling a possible overflow. in
addition, an overflow is now an explicit panic instead of a silent
overflow.

closes #52

PR #58, opened 2021-12-19
Protect overflow when computing expiration (part 2)

tatsuya6502:
This is a follow up PR for #56.

- Update `CacheBuilder::build_with_hasher` method to do the same duration checks to `build` methods.
- Update the API doc to explain panics.
- Not related to overflow checks but also tweaked the contents of the README.


tatsuya6502 (commented):
Merging.

PR #24, opened 2021-08-09
Support weight-based eviction

tatsuya6502:
- Add support for weight based (relative size based) eviction.
    - Implement "Aggregate Victims" (AV) algorithm introduced by this paper: [Lightweight Robust Size Aware Cache Management](https://arxiv.org/abs/2105.08770) by Gil Einziger, Ohad Eytan, Roy Friedman, Ben Manes
- Add support for unbound cache.

## Edit on Dec 31, 2021

Finished implementing above features on the all caches:

- `future::Cache`
- `sync::Cache`
- `sync::SegmentedCache`
- `unsync::Cache`

#### API:

- `Cache` and `SegmentedCache`
    - Changed the return type of `max_capacity()` from `usize` to `Option<usize>`.
    - Added `builder()` method.
- `CacheBuilder`
    - Made the max capacity optional.
    - Added `weigher` method to set the weigher closure.
        
#### Internal Changes (Highlighting major changes):

- `sync::BaseCache`
    - Updated `do_insert_with_hash` method to calculate the policy weight of the entry using the weigher closure.
- `sync::InnerCache`
    - Changed to track the total weighed size.
    - Completely rewritten `admit` and `find_cache_victim` methods to do the weight-based admission.
    - Added `evict_lru_entries` method to evict LRU entries when the weighed size exceeds the max capacity.
        - This is needed to handle a case when an existing admitted entry was updated with a larger policy weight.
- `sync::ValueEntry`
    - Updated to hold the policy weight of the entry via `EntryInfo` enum.
    - Added `EntryInfo` enum to reduce the number of `Arc` usages:
        - This may have positive impacts on memory consumption and performance.
        - But also this may have negative impacts on performance as it does dynamic dispatches.

### Notes

- We still need to evaluate the aggregated victim strategy to see if it will improve cache hit rate in various real-world workloads (e.g. by using Caffeine's simulator). See [this comment](https://github.com/moka-rs/moka/pull/24#commitcomment-54926612) for details.

PR #61, opened 2022-01-01
Lazily initialize and enable the LFU filter

tatsuya6502:
This is a follow-up PR for #24.

This PR makes these caches not to initialize the LFU filter (`FrequencySketch`) at the cache creation time, but initialize and enable it once the cache is half full (based on the weighted size).

The size of the `FrequencySketch` is determined by the followings:
- If the `weigher` is set, use the current number of the entries in the cache.
- If not, use the `max_capacity` of the cache, which is equals to the max number of entries.

This PR also changes the `max_capacity` from `usize` to `u64`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Version 0.7.0
 
+- **Breaking change**: The type of the `max_capacity` has been changed from `usize`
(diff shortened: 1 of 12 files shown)


tatsuya6502 (commented):
Merging.

PR #65, opened 2022-01-11
Fix memory leak when evicting or expiring an entry (v0.7.x)

### LINKED ISSUE #64: Memory leak
Hi team -- thanks for making `moka` available to the world!

I'm having some difficulty using it, for me the cache simply grow till it run out of memory.

Below is a minimum reproducible example. It grows to a few gigs of memory usage after a few minutes. Am I not incantating the cache properly? (thanks for your patience i'm a Rust n00b)

```rust
use moka::unsync::{Cache, CacheBuilder};
use ahash::RandomState;

use std::time::Duration;
use rand::Rng;

const CACHE_SIZE: u64 = 1_000_000;
const TTI: Duration = Duration::from_millis(10_000);

fn main() {
    let mut cache: Cache<(u64, u64, u64), (), RandomState> = CacheBuilder::new(CACHE_SIZE as usize)
        .initial_capacity(CACHE_SIZE as usize)
        .time_to_idle(TTI)
        .build_with_hasher(RandomState::default());

    let mut rng = rand::thread_rng();
    loop {
        let key = (rng.gen::<u64>(), rng.gen::<u64>(), rng.gen::<u64>());
        cache.insert(key, ());
    }
}
```

tatsuya6502:
This PR addresses the memory leak issue reported by #64.

The memory leak happens when evicting/expiring an entry or manually invalidating an entry. It should not happen when dropping a cache.

The leaking data is the key of an entry, last modified timestamp (`u64`) and last accessed timestamp (`u64`). It does not include the value of an entry.

## How to reproduce

- This issue can be reproduced by the code in the description of #64.
- Or by running the following commands. [Miri](https://github.com/rust-lang/miri) will detect similar issue in `Deque`'s unit tests:

```console
$ rustup update nightly
$ rustup +nightly component add miri
$ cargo +nightly miri test deque
```

```
running 4 tests
test common::deque::tests::basics ... ok
test common::deque::tests::drop ... ok
test common::deque::tests::iter ... ok
test common::deque::tests::next_node ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 39 filtered out

The following memory was leaked: alloc92503 (Rust heap, size: 1, align: 1) {
    61                                              │ a
}
alloc92586 (Rust heap, size: 48, align: 8) {
    0x00 │ 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 │ ................
    0x10 │ ╾0x23b1f7[a92503]<untagged> (8 ptr bytes)╼ 01 00 00 00 00 00 00 00 │ ╾──────╼........
    0x20 │ 01 00 00 00 00 00 00 00 01 __ __ __ __ __ __ __ │ .........░░░░░░░
}
alloc93007 (Rust heap, size: 1, align: 1) { ... }
alloc93087 (Rust heap, size: 48, align: 8) { ... }
alloc93620 (Rust heap, size: 1, align: 1) { ... }
alloc93700 (Rust heap, size: 48, align: 8) { ... }
alloc105252 (Rust heap, size: 1, align: 1) { ... }
alloc105323 (Rust heap, size: 48, align: 8) { ... }
alloc114558 (Rust heap, size: 1, align: 1) { ... }
alloc114629 (Rust heap, size: 48, align: 8) { ... }

error: the evaluated program leaked memory
```

## Cause

The root cause is forgetting to drop an unlinked `DeqNode` after calling an unsafe method `moka::common::Deque::unlink`.

## Fixes

- Added unsafe `unlink_and_drop` method to `Deque` and updated the callers of `unlink` method to call `unlink_and_drop`. As the name implies, `unlink_and _drop` will unlink the `DeqNode` and then drop it.
- Added Miri test to the CI to detect similar programming errors in future.

This PR is for the master branch (v0.7.x). There is another PR #66 with the same fixes for the maint-06 branch (v0.6.x).

## Affected Versions

- v0.1.0 to v0.5.4
- v0.6.0 to v0.6.3. (Fixed in v0.6.4)
- v0.7.0. (Fixed in v0.7.1)

## Notes

We are not yanking the affected versions at crates.io although this issue will have impact to almost all Moka users, and may lead out of memory error in some use cases. This is because memory leak is not considered a memory safety issue and does not lead undefined behavior (UB).

Instead, we are doing patch releases v0.6.4 and v0.7.1 with the fixes, so that existing v0.6 and v0.7 users will automatically move to one of these fixed versions when `cargo update` is executed.

* * *
Fixes #64.

### DIFF
--- a/.github/workflows/Miri.yml
+++ b/.github/workflows/Miri.yml
@@ -43,4 +43,5 @@ jobs:
       - name: Run Miri test (deque)
         uses: actions-rs/cargo@v1
         with:
-          command: miri test deque
+          command: miri
+          args: test deque
--- a/.github/workflows/Miri.yml
+++ b/.github/workflows/Miri.yml
@@ -1,0 +1,46 @@
+name: Miri tests
+
+on:
+  push:
+    paths-ignore:
+    - '.devcontainer/**'
+    - '.gitpod.yml'
+    - '.vscode/**'
+    - 'tests/**'
+  pull_request:
+    paths-ignore:
+    - '.devcontainer/**'
+    - '.gitpod.yml'
+    - '.vscode/**'
+    - 'tests/**'
+  schedule:
+    # Run against the last commit on the default branch on Friday at 9pm (UTC?)
+    - cron:  '0 21 * * 5'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Moka
+        uses: actions/checkout@v2
+
+      - name: Install Rust nightly toolchain with Miri
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
+          components: miri
+
+      - uses: Swatinem/rust-cache@v1
+
+      - name: cargo clean
+        uses: actions-rs/cargo@v1
+        with:
+          command: clean
+
+      - name: Run Miri test (deque)
+        uses: actions-rs/cargo@v1
+        with:
+          command: miri test deque
--- a/.github/workflows/Miri.yml
+++ b/.github/workflows/Miri.yml
@@ -29,7 +29,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: ${{ matrix.rust }}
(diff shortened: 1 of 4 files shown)


tatsuya6502 (commented):
Merging.

PR #66, opened 2022-01-11
Fix memory leak when evicting or expiring an entry (v0.6.x)

tatsuya6502:
This is the to #65 but for Moka v0.6.x.

Relates to #64.


tatsuya6502 (commented):
Merging.

PR #67, opened 2022-01-11
Update the minimum version of crossbeam-channel from v0.5.0 to v0.5.2

tatsuya6502:
This PR update `Cargo.toml` for switching the minimum version of crossbeam-channel from v0.5.0 to v0.5.2.

v0.5.2 addressed some stacked borrow violations found by Miri so it will be safer to depend on that version (or newer versions):

- https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-channel/CHANGELOG.md#version-052

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,10 @@ future = ["async-io", "async-lock", "futures-util"]
 atomic64 = []
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


tatsuya6502 (commented):
Merging.

Issue #72, opened 2022-01-18
Memory utilization grows significantly beyond max_capacity() under load


chriskyndrid:
Thank you for the excellent work on this interesting project. I've implemented Moka in a project I'm working on to utilize as a memory based cache. I'm passing a struct to bincode, encoding it, and storing it the cache, then running the reverse as needed. The system is a web services that utilizes Actix. In practice things are working fine. 

However, when I ran a load generating program `wrk` against a sample call, the memory utilization of the cache increases considerably and does not appear to respect the `max_capacity()` setting. It does not decrease after load generation completes and subsequent `wrk` tests continue to increase the memory footprint likely indefinitely. 

As an example:

```rust
lazy_static! {
    static ref MOKA_CACHE: MCache<Uuid, CacheItem> = MCache::builder()
        .weigher(|_key, value: &CacheItem| -> u32 {
            //let size = size_of_val(&*value.data) as u32;
            let size = value.data.len().try_into().unwrap_or(u32::MAX);
            size
        })
        .max_capacity(
                 256
                * 1024
                * 1024
        )
        .build();
}
```

**1:** 
At the start of the load generation:

Top Indicates around 15M of memory utilization:

`6488   erp_service_acco 0.0  00:01.19 29     0    50     15M`

**2:**
Executing a `wrk` request yields

```
wrk -d 10 -t 8 -c 100 http://localhost:8080/v1/account/customer/test/1

Running 10s test @ http://localhost:8080/v1/account/customer/test/1
  8 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.16ms    1.79ms  48.82ms   99.58%
    Req/Sec    11.23k     0.92k   13.45k    83.66%
  902635 requests in 10.10s, 314.20MB read
Requests/sec:  89370.39
Transfer/sec:     31.11MB
```

Top now indicates around **449M** of memory utilization

`6488   erp_service_acco 0.0  01:19.42 45     0    66     449M`

**3**
Executing a longer `wrk` request again

```
wrk -d 100 -t 8 -c 100 http://localhost:8080/v1/account/customer/test/1
Running 2m test @ http://localhost:8080/v1/account/customer/test/1
  8 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     4.97ms   12.33ms 306.14ms   92.16%
    Req/Sec     7.77k     4.35k   14.75k    71.28%
[...]

PR #75, opened 2022-02-01
Fix a bug that can create oversized frequency sketch when weigher is set

tatsuya6502:
Moka v0.7.0 introduced size-aware cache management. However, there is a bug in the logic to estimate the number of keys that the frequency sketch (the LFU filter) would manage, and can create oversized sketch up to two times bigger than necessary. This bug exists only for size-aware cache management.

This PR fixes the bug by trying not to increase the sketch size once estimated.

* * *
Relates to #72.

### DIFF
--- a/src/sync/base_cache.rs
+++ b/src/sync/base_cache.rs
@@ -384,7 +384,7 @@ where
             }
         }
         // Enable the frequency sketch.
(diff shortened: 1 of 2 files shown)


tatsuya6502 (commented):
Merging.

PR #76, opened 2022-02-01
Change `EntryInfo` from `enum` to `struct` to reduce memory utilization

tatsuya6502:
- To reduce memory utilization, change `EntryInfo` from `enum` to `struct`. (See #72)
- Remove `ValueEntryBuilder` and `CacheFeatures`. We used them to select an enum variant of `EntryInfo`.

* * *
Relates to #72.


tatsuya6502 (commented):
Merging.

PR #80, opened 2022-02-01
Replace some `std::sync::Arc` with `triomphe::Arc` to reduce memory utilization

tatsuya6502:
- Replace `Arc` for `Arc<ValueEntry>`, `Arc<EntryInfo>` and `Arc<value_initializer::Waiter>`.
- Add `triomphe` to the dependencies.

Note: Did not replace `Arc` for `Arc<K>` as this change leaks to the public APIs. e.g. `get(&Q)`.

* * *
Relates to #72.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -41,6 +41,7 @@
         "Tatsuya",
         "thiserror",
         "toolchain",
+        "triomphe",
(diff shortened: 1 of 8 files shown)


tatsuya6502 (commented):
Merging.

PR #82, opened 2022-02-02
Add unstable-debug-counters feature

tatsuya6502:
When this feature is enabled:
- `moka::GlobalDebugCounters::current` function is provided.
- `moka::future::Cache::debug_stats` method is provided.
- The eviction batch size is changed to 10,000.

* * *
Relates to #72.


tatsuya6502 (commented):
Merging.

PR #88, opened 2022-02-11
Improve performance on `get_or_insert_with`

tatsuya6502:
- Avoid to update a waiter in the concurrent hash table (cht) if it is already present.
- Overall refactoring in cht `bucket` and `bucket_array` modules.


tatsuya6502 (commented):

PR #89, opened 2022-02-11
Refactoring in `cht::*` modules

tatsuya6502:
- Move code in `cht::segment::map` module to `cht::segment` module.
- Move tests in `cht::test_util::tests` to `cht::segment` module.
- Re-enable tests in `cht::map::bucket` module.
- Fix Clippy warnings in cht tests.


tatsuya6502 (commented):

PR #90, opened 2022-02-12
Avoid to calculate the same hash twice in `get`, `insert`, `invalidate`, etc. methods

tatsuya6502:
Update crate-internal `cht` methods to take a hash value of the key so that they do not have to recalculate the hash from the key.


tatsuya6502 (commented):

PR #98, opened 2022-03-19
Prepare for v0.8.0 release

tatsuya6502:
- Bump the version to v0.8.0.
- Remove the integration test for actix-rt v1. And also remove actix-rt v1 from the dev dependencies.


tatsuya6502 (commented):
Merging.

PR #100, opened 2022-03-20
Update the minimum depending version of crossbeam-channel to v0.5.4

tatsuya6502:
crossbeam-channel v0.5.4 has a workaround for a bug in upstream (`rustc` or LLVM?) related to TLS access on AArch64 Linux: https://github.com/crossbeam-rs/crossbeam/pull/802

Ensure that Moka uses crossbeam-channel v0.5.4 or newer v0.5.x.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Version 0.8.0
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


tatsuya6502 (commented):
Merging.

PR #103, opened 2022-03-20
Bump `scheduled-thread-pool`

Milo123459:
This does two things, bumps `scheduled-thread-pool` to `0.2.5` (because anything below 0.2.4 would break, and it was using the minimal version) and adds a workflow to catch similar issues like this.


tatsuya6502 (approved):
> adds a workflow to catch similar issues like this.

Nice addition! Thank you!

PR #104, opened 2022-03-21
Raise the minimum depending version of skeptic to v0.13.5

tatsuya6502:
Need to raise the version to avoid compile errors with old lib and rand crates.


tatsuya6502 (commented):
Merging.

PR #114, opened 2022-04-09
Add `iter` method to create an iterator to `unsync`, `sync` and `future` caches

### LINKED ISSUE #112: Iterating cache
Hi @tatsuya6502 ,

I'm using moka cache (moka::future)...
```rust
use moka::future::{
    Cache,
    CacheBuilder,
};
...
lazy_static!{
    static ref WEB_CACHE: ResultCache =  CacheBuilder::new(10_000) // Max 10,000 elements
            .time_to_live(Duration::from_secs(15 * 60)) // Time to live (TTL): 15 minutes
            .time_to_idle(Duration::from_secs( 5 * 60)) // Time to idle (TTI):  5 minutes
            .build();  // Create the cache.
}
```

It is working well...

What is the best way of iterating through the elements in the list ?

I've looked through the code trying to find a way.
A "keys(&self) -> Vec<Key>" method would work..
Then I can iterate through the normal get...
Thanks...
   JR

tatsuya6502:
## Changes

This PR adds iterator visiting all key-value pairs in arbitrary order.

- Add the `iter` method, who returns an iterator, to the following caches:
    - Not-thread-safe cache:
        - `unsync::Cache`
    - Concurrent Caches:
        - `sync::Cache`
        - `sync::SegmentedCache`
        - `future::Cache`
- Implement `IntoIterator` to all the caches including experimental `dash::Cache`.

Fixes #112

### `unsync` cache's iterator

The `unsync::Cache` employs `std::collections::HashMap` as the central key-value storage:

- The iterator element type is `(&K, &V)`.
- No mutation on the cache is allowed while an iterator is alive.
    - Mutations includes both read operations like `get` and write operations like `insert`.

### `sync` and `future` caches' iterators

These concurrent caches employ `moka::cht::SegmentedHashMap` as the central key-value storage:

- The iterator element type is `(std::sync::Arc<K>, V)`, where `V` is a clone of the stored values.
- Concurrent mutations on the cache is allowed while an iterator is alive.

To allow concurrent mutations, iterator's `next` method does not guarantee the following:

- It does not guarantee to return a key-value pair (aka entry) if its key has been inserted to the cache _after_ the iterator was created.
    - Such an entry may or may not be returned depending on key's hash and timing.

and the `next` method guarantees the followings:

- It guarantees not to return the same entry more than once.
- It guarantees not to return an entry if it has been removed from the cache after the iterator was created.
    - Note: An entry can be removed by following reasons:
        - It has been manually invalidated.
        - It has been expired (e.g. time-to-live).
        - It has been evicted as the cache capacity exceeded.

### An implementation note

`sync` and `future` caches' iterators have some memory overheads:

It creates a snapshot of keys (~~`Vec<std::sync::Weak<K>>`~~ `Vec<std::sync::Arc<K>>`) of an internal segment of `cht::SegmentedHashMap`. It uses some amount of heap memory depending on the number of keys in the segment. The number of segments of `cht::SegmentedHashMap` per cache is currently hard-coded to 64 (since Moka v0.1.0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Moka &mdash; Change Log
 
+## Version 0.8.2
+
+### Added
+
+- Add iterator to the following caches: ([#114][gh-pull-0114])
+    - `sync::Cache`
+    - `sync::SegmentedCache`
+    - `future::Cache`
+    - `unsync::Cache`
+- Implement `IntoIterator` to the all caches (including experimental `dash::Cache`)
+  ([#114][gh-pull-0114])
+
+
 ## Version 0.8.1
 
 ### Added
@@ -287,6 +300,7 @@ The minimum supported Rust version (MSRV) is now 1.51.0 (2021-03-25).
 [gh-issue-0038]: https://github.com/moka-rs/moka/issues/38/
 [gh-issue-0031]: https://github.com/moka-rs/moka/issues/31/
 
+[gh-pull-0114]: https://github.com/moka-rs/moka/pull/114/
 [gh-pull-0105]: https://github.com/moka-rs/moka/pull/105/
 [gh-pull-0104]: https://github.com/moka-rs/moka/pull/104/
 [gh-pull-0103]: https://github.com/moka-rs/moka/pull/103/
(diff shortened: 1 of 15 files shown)


tatsuya6502 (commented):
Merging.

PR #116, opened 2022-04-11
Fix the `dash::Cache` iterator not to return expired entries

tatsuya6502:
This PR fixes a bug that the `dash::Cache` iterator can return expired entries.

- Now the `next` method of `dash::Cache` iterator checks whether next entry is expired or not and skip it when expired.
- Update the unit test cases to detect this bug.


tatsuya6502 (commented):
Merging.

PR #120, opened 2022-05-11
CI: Fix build error with MSRV (hashbrown >= v0.12 requires Rust 2021 edition)

tatsuya6502:
In MSRV tests, avoid hashbrown >= v0.12, which requires Rust 2021 edition.


tatsuya6502 (commented):
All CI passed. Merging.

PR #124, opened 2022-05-16
Add `get_with_if` method to `sync` and `future` caches

### LINKED ISSUE #123: [feature request] add `Cache::get_with_if(&self, key: &K, fallback: Fut, cond: Fn(&V) -> bool) -> V` method
I'm building a cache whose entries have their own standalone TTL. However `moka` supports global TTL and IDLE only. So I'm here to propose a new method `Cache::get_with_if(&self, key: &K, fallback: Fut, cond: Fn(&V) -> bool) -> V`

The function works like `Cache::get_with(&self, key: &K, init: Fut) -> V`, but when the key does not exist or `cond(&V)` is true, the `init` clojure will be called to initialize the value.

tatsuya6502:
This adds `get_with_if` method requested by #123 to the following caches:

- `sync::Cache`
- `sync::SegmentedCache`
- `future::Cache`
 
* * *
Fixes #123

### DIFF
--- a/src/future/cache.rs
+++ b/src/future/cache.rs
@@ -497,7 +497,19 @@ where
     pub async fn get_with(&self, key: K, init: impl Future<Output = V>) -> V {
         let hash = self.base.hash(&key);
         let key = Arc::new(key);
-        self.get_or_insert_with_hash_and_fun(key, hash, init).await
+        let replace_if = None as Option<fn(&V) -> bool>;
+        self.get_or_insert_with_hash_and_fun(key, hash, init, replace_if)
+            .await
+    }
+
+    pub async fn get_with_if<F>(&self, key: K, init: impl Future<Output = V>, replace_if: F) -> V
+    where
+        F: Fn(&V) -> bool + Send + Sync + 'static,
+    {
+        let hash = self.base.hash(&key);
+        let key = Arc::new(key);
+        self.get_or_insert_with_hash_and_fun(key, hash, init, Some(replace_if))
+            .await
     }
 
     /// Try to ensure the value of the key exists by inserting an `Ok` output of the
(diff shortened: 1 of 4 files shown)


src/future/cache.rs
```diff
@@ -830,9 +844,12 @@ where
         key: Arc<K>,
         hash: u64,
         init: impl Future<Output = V>,
+        replace_if: Option<impl Fn(&V) -> bool>,
```
ClSlaid, line 847: looks like this function is relied on by many methods of the cache, maybe the performance loss on adding a new `match` assertion should be benchmarked.

ClSlaid (commented):
Clean code, wealthy coverage of tests, and quick reaction to issues, this is the best rust cache crate I've ever seen.

src/future/cache.rs
tatsuya6502: Thank you for the suggestion. The `get_with` method will be affected by the `match` condition change although it should have relatively smaller performance impact than calling `self.base.get_with_hash(&key, hash)` (which already exists in older versions).

Just for sure, I created a micro benchmark using [Criterion](https://crates.io/crates/criterion) to compare performance of `get_with` method between previous Moka v0.8.3 and the current version. 

- https://gitlab.com/moka-labs/moka-gh-issue-123/moka-microbench/-/blob/main/benches/future_cache_benchmark.rs

I run it on my Mac mini M1 and Criterion reported "No change in performance detected."

```console
## Moka v0.8.3 (baseline)
$ cargo bench --features moka083
...

get_with                time:   [2.2915 us 2.2938 us 2.2964 us]                      
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe

## Moka with current change
$ cargo bench --features moka084
...

get_with                time:   [2.3039 us 2.3064 us 2.3093 us]                      
                        change: [-0.1492% +0.2773% +0.6093%] (p = 0.16 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe
```

PR #127, opened 2022-05-19
Remove ptr-int transmute by converting UnsafeWeakPointer from usize to *mut T

saethlin:
This transmute from a `Weak<T>` to a `usize` is shockingly common in both the ecosystem and the standard library. It's doubly bad because `Weak<T>` is a `repr(Rust)` type, so without manually inspecting its layout at runtime it is a mistake to assume it has any particular layout. But primarily, this is a round-trip of a pointer through an integer type via transmute, and it's pretty clear Rust cannot have a sound memory model that permits these alongside the usual suite of compiler optimizations (casts are different, they're annoying but possible to support).

If you're looking for other examples of this approach, `rayon` does a very similar thing to get pointers between threads: https://github.com/rayon-rs/rayon/blob/5b6adbbf5c6944ec42b54530820ee8c0ad0e35a9/src/lib.rs#L123-L147

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ unstable-debug-counters = ["future"]
 
 [dependencies]
 crossbeam-channel = "0.5.4"
-crossbeam-epoch = "0.9"
+crossbeam-epoch = "0.8.2"
 crossbeam-utils = "0.8"
 num_cpus = "1.13"
 once_cell = "1.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ unstable-debug-counters = ["future"]
 
 [dependencies]
 crossbeam-channel = "0.5.4"
-crossbeam-epoch = "0.8.2"
+crossbeam-epoch = "0.9"
 crossbeam-utils = "0.8"
 num_cpus = "1.13"
 once_cell = "1.7"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -39,7 +39,7 @@ unstable-debug-counters = ["future"]
 
 [dependencies]
 crossbeam-channel = "0.5.4"
-crossbeam-epoch = "0.8.2"
+crossbeam-epoch = "0.9"
```
tatsuya6502, line 42: Can you please revert this change?

Upgrading crossbeam-epoch needs extensive testing to ensure #34 will not get worse, and I am not ready for it now.

tatsuya6502 (changes requested):
Thank you for reviewing and fixing the unsafe codes. Greatly appreciated!

I left a comment about reverting the crossbeam-epoch version. All other parts look great to me.

Cargo.toml
saethlin: That's reasonable. I just wanted to bump this because the transmutes are detected by the same Miri flag (`-Zmiri-check-number-validity`) which also detects this crossbeam issue: https://github.com/crossbeam-rs/crossbeam/pull/779

tatsuya6502: Thanks. Yeah. I know I should upgrade crossbeam-epoch soon to get those important fixes/improvements. I will try to find some time for it.

tatsuya6502 (approved):
Thanks. It looks good to me!

PR #129, opened 2022-05-20
Increase the num segments of the waiters hash table from 16 to 64

tatsuya6502:
This will reduce the chance of issue #34 occurring.


tatsuya6502 (commented):
Merging.

PR #133, opened 2022-05-21
Remove unnecessary `K: Clone` bound from some caches when they are `Clone`

### LINKED ISSUE #131: `<Cache as Clone>` has overly-restrictive trait bounds
The various `Cache` types derive `Clone`, which means they require `where K: Clone, V: Clone, S: Clone`. This is overly restrictive. Since the underlying `BaseCache` on each type has no trait bounds on their `Clone` impls, the `Cache`s don't need them either.

Besides being annoying, this also means that the cache key needs to be cloneable in order to clone the cache, even though this is not otherwise a requirement for cache keys.

tatsuya6502:
Remove unnecessary `K: Clone` bound from the following caches when they are `Clone`:

- `sync::Cache`
- `future::Cache`
- `dash::Cache`

* * *
Fixes #131

### DIFF
--- a/.github/workflows/Skeptic.yml
+++ b/.github/workflows/Skeptic.yml
@@ -62,7 +62,7 @@ jobs:
         env:
           RUSTFLAGS: '--cfg skeptic'
 
-      - name: Run UI tests (future and dash features, trybuild)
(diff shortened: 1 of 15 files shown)


tatsuya6502 (commented):
Merging.

PR #138, opened 2022-05-25
Add `Debug` impl to the following caches

### LINKED ISSUE #132: `Cache` missing `Debug` impl
The `Cache` types are all missing `Debug` impls, which is annoying. I'm not sure what offhand such an impl should show, but perhaps some basic stats about the cache if that's easy to gather.

tatsuya6502:
- `sync::Cache`
- `sync::SegmentedCache`
- `future::Cache`
- `unsync::Cache`

Note that `dash::Cache` already has `Debug` impl.

* * * 
Fixes #132

### DIFF
--- a/src/future/cache.rs
+++ b/src/future/cache.rs
@@ -20,6 +20,7 @@ use std::{
     any::TypeId,
     borrow::Borrow,
     collections::hash_map::RandomState,
+    fmt,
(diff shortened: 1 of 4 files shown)


tatsuya6502 (commented):
Merging.

Issue #140, opened 2022-05-26
Are there certain dependencies that can be removed?


Milo123459:
Currently, moka has a lot of dependencies. I was wondering if there were any that could be replaced. I'd love to work on this.

tatsuya6502:
Hi. Thank you for the offer. Well, all these dependencies are in-use, of course. However if you only use `moka::dash::Cache` in `serenity`, you could remove the following dependencies from `moka`:

- `crossbeam-epoch`
- `uuid`
- `thiserror` (but `serenity` already depends on it)

To do that, you will need the followings:

1. Add a crate feature called `sync` to `Cargo.toml`, and add it to the default features.
2. In `Cargo.toml`, make the following dependencies optional, and enable them only when `sync` or `future` feature is specified:
    - `crossbeam-epoch`
    - `uuid`
    - `thiserror`
3. Make the following modules enabled only when `sync` feature is specified: (Use `#[cfg(feature="sync")]`)
    - `sync::builder`
    - `sync::cache`
    - `sync::segments`
    - `sync::value_initializer`
4. Make the following modules enabled only when `sync` or `future` feature is specified: (Use `#[cfg(any(feature="sync", feature="future"))]`)
    - `sync`
    - `cht`
    - `common::error`

(I have not tried above by myself. So there may be additional changes required)

Once done, and a new version of `moka` is published to crate.io, you will update `moka` dependency in `serenity` to the followings:

**`serenity`'s `Cargo.toml`**

```toml:Cargo.toml
[dependencies.moka]
version = "0.9"
default-features = false
features = ["dash", "atomic64", "quanta"]
optional = true
```

Also I found that `scheduled-thread-pool v0.2.5` uses an older version of `parking_lot` (`v0.11.2`). `serenity` already depends on `parking_lot` but a newer version (`v0.12.0`). I see the `master` branch of sfackler/scheduled-thread-pool has it already upgraded to `v0.12` via https://github.com/sfackler/scheduled-thread-pool/pull/15, but no new version of `scheduled-thread-pool` has been published to crates.io.

So could you please open a GitHub issue to [sfackler/scheduled-thread-pool](https://github.com/sfackler/scheduled-thread-pool) and ask them to publish a new version to crates.io?

Thanks!

Milo123459:
https://github.com/moka-rs/moka/pull/141 made the pr!

tatsuya6502:
Thanks for making the PR #141. Hmm, there are so many compile errors.

[...]

PR #142, opened 2022-05-29
Refactoring: Reorganize internal modules

tatsuya6502:
## Changes

This PR reorganizes internal modules to make it easier to add a crate feature `sync` later for #140:

- Add `common::concurrent` module and move the modules shared by `sync`, `future` and `dash` caches there.
- Add `sync_base` module and move the modules shared by `sync` and `future` caches there.
- Add `common::concurrent::constants` module and move `const`s shared by `sync::base_cache` and `dash::base_cache`, etc. there.

Also it bumps the Moka version to v0.8.6.

## Additional Tasks

- [x] Generate the doc and ensure these internal modules `common::concurrent` etc. are not exposed in the doc.


tatsuya6502:
There is a hyperlink issue in the generated rustdoc:

- When `future` feature is enabled, the doc pages of `sync::Cache` and `sync::SegmentedCache` have links to `future::Iter` and `future::PredicateId`, instead of `sync::Iter` and `sync::PredicateId`.
- When `future` feature is not enabled, the same doc pages have links to the correct ones.
- In either cases, there are the doc pages for `sync::Iter` and `sync::PredicateId`, so it is only a hyperlink issue.

I believe this is because `Iter` and `PredicateId` are defined in `sync_base` module which is not public. And both `future` and `sync` modules re-export them.

tatsuya6502:
> There is a hyperlink issue in the generated rustdoc:

I am going to remove `future`s re-exports. This will fix the issue. The current versions (v0.8.x) do not have `future::Iter` and `future::PredicateId` so I should not have added the re-export at the first place.

tatsuya6502 (commented):
Merging.

PR #145, opened 2022-05-30
Add support for eviction listener

### LINKED ISSUE #128: Feature request: Add an `eviction_callback` parameter to the cache builder
Hey there, first and foremost thanks for moka!

I currently have this use case where I have other data structures I need to keep in sync with a moka cache. One way I though of doing this is the cache builder could support a new `eviction_callback` method which takes a closure that is called every time there is a cache eviction, with ideally both the key and value as parameters to the closure. Do you think that's feasible and if so, is this something you'd be interested in this being added to moka?

I envision something like:

```rust
pub fn eviction_callback<F>(&self, predicate: F)
where
    F: Fn(&K, &V) + Send + Sync + 'static,
```

or

```rust
pub fn eviction_callback<F>(&self, predicate: F)
where
    F: Fn(K, &V) + Send + Sync + 'static,
    K: Copy
```

tatsuya6502:
This PR adds a support for the eviction listener to the following caches:

- `sync::Cache`
- `sync::SegmentedCache`
- `future::Cache`

The eviction listener is a closure with the following signature and can be used (for example) to keep other data structures in-sync with the cache.

```rust
impl Fn(Arc<K>, V, RemovalCause) + Send + Sync + 'static
where
    V: Clone,
```

and the enum `RemovalCause` (`moka::notification::RemovalCause`) is defined as the followings:

```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum RemovalCause {
    /// The entry's expiration timestamp has passed.
    Expired,
    /// The entry was manually removed by the user.
    Explicit,
    /// The entry itself was not actually removed, but its value was replaced by
    /// the user.
    Replaced,
    /// The entry was evicted due to size constraints.
    Size,
}
```

The enum `DeliveryMode` specifies how and when an eviction notification should be delivered to an eviction listener.

```rust
#[derive(Copy, Clone, Debug, PartialEq, Eq)]
pub enum DeliveryMode {
    /// With this mode, a notification should be delivered to the listener
    /// immediately after an entry was evicted. It also guarantees that eviction
    /// notifications and cache write operations such and `insert`, `get_with` and
    /// `invalidate` for a given cache key are ordered by the time when they
    /// occurred.
    ///
    /// To guarantee the order, cache maintains key-level lock, which will reduce
    /// concurrent write performance.
    ///
    /// Use this mode when the order is more import than the write performance.
    Immediate,
    /// With this mode, a notification will be delivered to the listener some time
    /// after an entry was evicted. Therefore, it does not preserve the order of
    /// eviction notifications and write operations.
    ///
    /// On the other hand, cache does not maintain key-level lock, so there will be
    /// no overhead on write performance.
    ///
    /// Use this mode when write performance is more important than preserving the
    /// order of eviction notifications and write operations.
    Queued,
}
```

Currently, `sync` caches support both modes and `Immediate` mode is the default. But `future` cache supports only `Queued` mode.

* * *
Fixes #128


tatsuya6502 (commented):
Merging.

PR #149, opened 2022-06-17
Fix the CI for MSRV by pinning the versions of indexmap and native-tls

tatsuya6502:
- To avoid compile errors when testing MSRV (1.51.0), run the following commands for MSRV CI to pin the versions of indexmap and native-tls crates:
    - `cargo update -p indexmap --precise 1.8.2`
    - `cargo update -p native-tls --precise 0.2.8`
- Run `cargo tree --features 'future, dash'` on the main CI.


tatsuya6502 (commented):
Merging.

PR #150, opened 2022-06-18
Fix async integration tests

tatsuya6502:
The all async integration tests under the tests folder were not running as they were ignored by the test runner. Fix the issue by correcting the attributes on the modules and functions.


tatsuya6502 (commented):
Merging.

PR #158, opened 2022-07-02
Fix CI failures for Rust 1.62 and Clippy beta 0.1.63

tatsuya6502:
- Update `check_struct_size` of `EntryInfo` struct as the size of the struct changed in Rust 1.62 for 32 bit platforms.
    - `EntryInfo` on some 32 bit platform has a field with `RwLock`.
- Fix Clippy warnings. (Clippy beta 0.1.63, 6c1f1428998 2022-06-28)


tatsuya6502 (commented):
Merging.

PR #159, opened 2022-07-02
Add `logging` feature to enable optional `log` crate dependency

tatsuya6502:
This PR adds `logging` feature, which enables optional dependency to `log` crate. (This feature will be used by the eviction listener #145)


tatsuya6502 (commented):
Merging.

PR #164, opened 2022-07-09
Fix the doc example for eviction listener with the immediate mode

tatsuya6502:
It was using the queued mode which was incorrect.


tatsuya6502 (commented):
Merging.

PR #170, opened 2022-07-23
Fix a test `entry_info::check_struct_size` for macOS

tatsuya6502:
Fix a test `entry_info::check_struct_size` for macOS. Also skip it on non-Linux or non-macOS platforms.


tatsuya6502 (commented):
Merging.

PR #175, opened 2022-08-29
Follow up the `js` feature

tatsuya6502:
Move `getrandom` back to under the dev-dependencies.

Fixes the unit test broken by #173.


tatsuya6502 (commented):
Merging.

PR #181, opened 2022-09-24
CI: Pin once_cell to v1.14.0 for the MSRV

tatsuya6502:
once_cell v1.15.0 or newer requires Rust 2021 edition, which is not supported by the MSRV.


tatsuya6502 (commented):
Merging.

PR #186, opened 2022-10-22
CI: Update the expected memsize of the EntryInfo struct for Rust 1.66 nightly

tatsuya6502:
- Update the expected memsize of the `EntryInfo` struct for Rust 1.66 nightly.
- Add rustc_version crate to the build-dependencies to check the Rust version. (Only enabled when `RUSTFLAGS='--cfg rustver'` is set)
- To workaround the following Cargo issue with cross compiling, force enable rustc_version crate when testing Linux corss targets:
    - https://github.com/rust-lang/cargo/issues/4423


tatsuya6502 (commented):
Merging.

Issue #184, opened 2022-10-20
Add new api similar to `try_get_with`, but return `Option` rather than `Result`?


LMJW:
Thanks for the great work on the library!

I am using the library in one of the project that I am working on and I feel some apis similar to [try_get_with](https://docs.rs/moka/latest/moka/sync/struct.Cache.html#method.try_get_with), but instead of the function return `Result<V,E>`, having Function returns `Option<V>` would be more "ergonomics". 

Just want to know if you are have any objections of adding such APIs. Maybe something like "optional_get_with"? Happy to contribute if you are okay with it. Thanks

tatsuya6502:
Thank you for offering contribution! I like the idea of adding such API. The name `optional_get_with` will be good enough I think.

Actually, when I saw your idea, I wondered if we could make `try_get_with` more generic like the `?` operator. `?` operator not only accepts `Result` but also `Option` and `ControlFlow`. And once [try_trait_v2](https://github.com/rust-lang/rfcs/pull/3058) feature is stabilized, it will accept any type that implements `Try` trait.

I played around with my idea and came up with the following code using `Try` trait. It almost worked, but I could not find a way to solve the "Problem" in the code. So I think adding `optional_get_with` is a good idea for now.

```rust
// As of today (late 2022), Rust Nightly and the following feature is required:
#![feature(try_trait_v2)]

use std::{
    convert::Infallible,
    ops::{ControlFlow, Try},
    sync::Arc,
};

#[derive(Default)]
struct Cache<K, V>(std::collections::HashMap<K, V>);

impl<K, V> Cache<K, V> {
    // try_get_with that is generic over Result, Option, ControlFlow, etc.
    fn try_get_with<F, R>(&mut self, _key: K, init: F) -> R
    where
        F: FnOnce() -> R,
        R: Try<Output = V>,
    {
        match init().branch() {
            ControlFlow::Continue(v) => R::from_output(v),
            ControlFlow::Break(r) => R::from_residual(r),
        }
    }

    // Problem: Cannot encode the following requirements into the type.
    //
    // When F's concrete return type is Result<V, E> or ControlFlow<E, V>,
    // - E: Send + Sync + 'static is required.
    // - try_get_with's return type must be Result<V, Arc<E>> or ControlFlow<Arc<E>, V>
    //   instead of Result<K, E> or ControlFlow<E, K>.


    // This will emulate today's try_get_with.
    #[allow(dead_code)]
[...]

PR #187, opened 2022-10-23
Add new api `optionally_get_with`

LMJW:
This is the inital PR for issue #184 

I was thinking to use TypeId with `()` type for the Error type when return `InitResult`, but not sure if this will have some conflict with other parts of the code as I saw `()` is use somewhere else. So I created internal structure `OptionallyNone` to represent this error type. 

I also noticing the guarantee of call once only if the init closure return `Ok/Some`. If it returns `Err/None`, it is not gurantee to have once call. I think this definitely make sense, perhaps it is good to update the docummentation? Thanks

Let me know if there's any comments or suggestions. Thanks

### DIFF
--- a/src/future/value_initializer.rs
+++ b/src/future/value_initializer.rs
@@ -172,8 +172,7 @@ where
 
         // This closure will be called after the init closure has returned a value.
         // It will convert the returned value (from init) into an InitResult.
-        let post_init = |key, value: Option<V>, mut guard: WaiterGuard<'_, K, V, S>| match value
-        {
+        let post_init = |key, value: Option<V>, mut guard: WaiterGuard<'_, K, V, S>| match value {
             Some(value) => {
                 guard.set_waiter_value(WaiterValue::Ready(Ok(value.clone())));
                 InitResult::Initialized(value)
@@ -182,7 +181,7 @@ where
                 // `value` can be either `Some` or `None`. For `None` case, without
                 // change the existing API too much, we will need to convert `None`
                 // to Arc<E> here. `Infalliable` could not be instantiated. So it
-                // might be good to use an empty struct to indicate the error type. 
+                // might be good to use an empty struct to indicate the error type.
                 let err: ErrorObject = Arc::new(OptionallyNone);
                 guard.set_waiter_value(WaiterValue::Ready(Err(Arc::clone(&err))));
                 self.remove_waiter(key, type_id);
--- a/src/future/value_initializer.rs
+++ b/src/future/value_initializer.rs
@@ -7,6 +7,8 @@ use std::{
 };
 use triomphe::Arc as TrioArc;
 
+use super::OptionallyNone;
+
 const WAITER_MAP_NUM_SEGMENTS: usize = 64;
 
 type ErrorObject = Arc<dyn Any + Send + Sync + 'static>;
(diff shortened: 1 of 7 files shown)


LMJW:
@tatsuya6502 Feel free to take a look when you have time :)

LMJW:
Hmm, looks like the failed CI is not related to the new API changes? https://github.com/moka-rs/moka/blob/master/src/common/concurrent/entry_info.rs#L145

tatsuya6502:
Thank you very much for the PR. I will review it when I have some spare time. (hopefully tonight, UTC+0800).

You are right. The failed CI is not related to this PR, so do not worry about it. It is already failing on the master branch when Rust 1.66 nightly is used. I am trying to fix it by #186.

tatsuya6502:
> I also noticing the guarantee of call once only if the init closure return `Ok/Some`. If it returns `Err/None`, it is not gurantee to have once call. I think this definitely make sense, perhaps it is good to update the docummentation? Thanks

Thank you for the suggestion. Please feel free to edit the documentation and submit a PR if you can make it better. That would be great help. I am not a native English speaker and having a bit struggle to write clear, concise and consistent English.

https://docs.rs/moka/latest/moka/sync/struct.Cache.html#method.try_get_with
> This method prevents to evaluate the init closure multiple times on the same key even if the method is concurrently called by many threads; only one of the calls evaluates its closure ...

I think the word "even if" is a wrong choice here? It would imply "always". But what the cache actually does is _coalescing concurrent queries_.

* Concurrent queries are coalesced into one evaluation of the init closure.
* Sequential queries are not coalesced if they do not overlap.
* The entry will not be in the cache (so the init closure should be evaluated) if:
    * The entry was never inserted:
        * including: The init closure of previous query returned `Err/None`.
    * The entry was inserted but then evicted by size a constraint or expiration (e.g. time-to-live).

src/future/value_initializer.rs
```diff
@@ -156,6 +158,40 @@ where
...
+            Some(value) => {
+                guard.set_waiter_value(WaiterValue::Ready(Ok(value.clone())));
+                InitResult::Initialized(value)
+            }
+            None => {
+                // `value` can be either `Some` or `None`. For `None` case, without
+                // change the existing API too much, we will need to convert `None`
+                // to Arc<E> here. `Infalliable` could not be instantiated. So it
+                // might be good to use an empty struct to indicate the error type.
+                let err: ErrorObject = Arc::new(OptionallyNone);
```
tatsuya6502, line 185: > `Infalliable` could not be instantiated.

Thanks. You are right. `Infallible` is a unit enum and not struct, so it could not be instantiated.

tatsuya6502 (approved):
It looks good to me! Thank you for adding the unit tests and documentation.

tatsuya6502:
@LMJW 

> I also noticing the guarantee of call once only if the init closure return `Ok/Some`. If it returns `Err/None`, it is not gurantee to have once call. I think this definitely make sense, perhaps it is good to update the docummentation?

FYI, I released [v0.9.6](https://crates.io/crates/moka/0.9.6) with updated document. I hope it will be more clear about the guarantee.

I also found that there are small chances for race conditions in concurrent calls of `get_with`, etc., which can make the `init` closure to be evaluated more than once. I believe I fixed it for v0.9.6 via #195.

## The Documentation Update

### Before

https://docs.rs/moka/0.9.5/moka/sync/struct.Cache.html#method.optionally_get_with

> Try to ensure the value of the key exists by inserting an `Some` result of the init closure if not exist, and returns a _clone_ of the value or `None` returned by the closure.
> 
> This method prevents to evaluate the init closure multiple times on the same key even if the method is concurrently called by many threads; only one of the calls evaluates its closure (as long as these closures return the same Option type), and other calls wait for that closure to complete.

### After

https://docs.rs/moka/0.9.6/moka/sync/struct.Cache.html#method.optionally_get_with

> Returns a _clone_ of the value corresponding to the key. If the value does not exist, evaluates the `init` closure, and inserts the value if `Some(value)` was returned. If `None` was returned from the closure, this method does not insert a value and returns `None`.
>
> **Concurrent calls on the same key**
> 
> This method guarantees that concurrent calls on the same not-existing key are coalesced into one evaluation of the `init` closure. Only one of the calls evaluates its closure, and other calls wait for that closure to complete.
> 
> The following code snippet demonstrates this behavior:

LMJW:
Hi @tatsuya6502 Thanks for letting me know. The change looks great!

PR #193, opened 2022-11-01
Add the "entry" and "entry_by_ref" APIs

tatsuya6502:
Add `entry` and `entry_by_ref` APIs to `sync` and `future` caches. They return a struct `OwnedKeyEntrySelector` or `RefKeyEntrySelector` respectively. You can use their methods like `or_insert` to select or insert an entry of a cache.

These methods on the entry selectors return an `Entry`, a snapshot of key and value, carrying additional information `is_fresh` to indicates if the value was freshly computed or cached.

```rust
use moka::sync::Cache;

let cache: Cache<String, u32> = Cache::new(100);
let key = "key1".to_string();

let entry = cache.entry_by_ref(&key).or_insert(3);
assert!(entry.is_fresh());
assert_eq!(entry.key(), &key);
assert_eq!(entry.into_value(), 3);

let entry = cache.entry(key).or_insert(6);
// Not fresh because the value was already in the cache.
assert!(!entry.is_fresh());
assert_eq!(entry.into_value(), 3);
```

`OwnedKeyEntrySelector` and `RefKeyEntrySelector` structs provide the following methods:

- `or_default`
- `or_insert`
- `or_insert_with`
- `or_insert_with_if`
- `or_optionally_insert_with`
- `or_try_insert_with`

Also deprecate `get_with_if` methods in `sync` and `future` caches. The same functionality is provided by the entry selector's `or_insert_with_if` method.

### DIFF
--- a/src/future/cache.rs
+++ b/src/future/cache.rs
@@ -1,6 +1,7 @@
 use super::{
     value_initializer::{InitResult, ValueInitializer},
-    CacheBuilder, ConcurrentCacheExt, Iter, PredicateId,
+    CacheBuilder, ConcurrentCacheExt, Iter, OwnedKeyEntrySelector, PredicateId,
+    RefKeyEntrySelector,
 };
 use crate::{
     common::{
@@ -13,7 +14,7 @@ use crate::{
     },
     notification::{self, EvictionListener},
     sync_base::base_cache::{BaseCache, HouseKeeperArc},
-    Policy, PredicateError,
+    Entry, Policy, PredicateError,
 };
 
 #[cfg(feature = "unstable-debug-counters")]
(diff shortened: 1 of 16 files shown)


src/future/cache.rs
```diff
@@ -1421,170 +1502,238 @@ where
...
+        F: Future<Output = Option<V>>,
     {
         use futures_util::FutureExt;
 
         let get = || {
             let ignore_if = None as Option<&mut fn(&V) -> bool>;
             self.base
-                .get_with_hash_but_no_recording(&key, hash, ignore_if)
+                .get_with_hash_but_no_recording(&key, hash, ignore_if, false)
+                .map(Entry::into_value)
```
tatsuya6502, line 1632: Perhaps this method does not need `need_key` param and should return `Option<V>`?

tatsuya6502 (commented):
Left some comments.

tatsuya6502 (commented):
Merging.

PR #195, opened 2022-11-04
Prevent race condition in `get_with` method to avoid evaluating `init` closure/future multiple times

tatsuya6502:
In the current implementation (up to v0.9.5), there is a small chance that the following methods to evaluate the `init` closure/function more than once in concurrent calls:

- `get_with`
- `get_with_if`
- `optionally_get_with`
- `try_get_with`

It will happen when a call to one of these methods is made when another call to the same method on the same key is about to complete.

New test cases in he following files will reproduce the issue:

- `tests/entry_api_sync.rs`
- `tests/entry_api_tokio.rs`
- `tests/entry_api_actix_rt2.rs`
- `tests/entry_api_async_std.rs`

This PR fixes the issue by doing a double check if the value still does not exist after the lock is acquired and also extending the lock as needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,6 +88,7 @@ anyhow = "1.0.19"
 async-std = { version = "1.11", features = ["attributes"] }
 env_logger = "0.9"
 getrandom = "0.2"
+paste = "1.0.9"
 reqwest = "0.11.11"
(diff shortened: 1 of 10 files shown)


tatsuya6502 (commented):
Merging.

PR #200, opened 2022-11-09
Move `unsync` and experimental `dash` caches to a new crate called mini-moka

tatsuya6502:
This PR marks `unsync` cache deprecated and completely removes experimental `dash` cache from `moka` crate. They have been moved to a new crate called [`mini-moka`](https://github.com/moka-rs/mini-moka).

diff

### `dash` cache

- An experimental `dash` cache was added to `moka` at v0.8.0.
    - It was an alternative implementation to `sync` cache.
    - At the time, `sync` cache had an issue #34.
    - That issue was resolved at v0.9.2.
- `dash` cache has a smaller but duplicate feature set to `sync` cache.
    - Now the issue was resolved, there are not many reasons to keep `dash` cache in `moka` crate.
- However, there are some users who are happy with `dash` cash as it provides a smaller but enough feature set for their use cases.
    - So instead of removing `dash` cache, it would be nice if we move it to a new crate `mini-moka` that will provide simpler and lighter cache implementations.

### `unsync` cache

- `unsync` cache seems not popular and has a smaller feature set to `sync` and `future` cache.
   - If we move `dash` cache to `mini-moka`, it would make sense to move `unsync` cache to `mini-moka` too.

### Why removing `dash` cache right now from `moka` while keeping `unsync` cache as deprecated?

- `dash` cache shares some base modules with `sync` and `future` caches.
    - Since `dash` provides different (smaller) feature set than others, it is becoming difficult to maintain the base modules.
    - So removing it now will make the maintenance and development woks much easier.
- If `moka` user enables `dash` feature, it will fail to compile and _show a message to introduce `mini-moka` crate_.

- On the other hand, `unsync` cache shares very little part of the base modules with other caches.
    - It is not making maintenance problems now.
- Also, `unsync` cache is always enabled, and there is no crate feature like `dash` for it.
    - Thus, there is no way to show the message to introduce `mini-moka` crate if we completely remove it from `moka`.
    - So for now, we only mark it deprecated, and make the deprecation warning to introduce `mini-moka` crate.
    - We will remove `unsync` cache later at v0.11.0.


tatsuya6502 (commented):
Merging.

PR #202, opened 2022-11-19
Fix the CI for MSRV 1.51

tatsuya6502:
When testing the MSRV, downgrade `reqwest` to v0.11.12, otherwise we cannot downgrade `native-tls` to v0.2.8.


tatsuya6502 (commented):
Merging.

Issue #203, opened 2022-11-19
moka is too polymorphic (cargo-llvm-lines)


Swatinem:
I was playing around with https://github.com/dtolnay/cargo-llvm-lines recently, and found out that `moka` does generate a ton of llvm IR due to too much internal generics.

This is suboptimal for a couple of reasons:
* it generates multiple copied of functions that have to be compiled, hurting compile times.
* may lead to binary bloat
* may even hurt performance as the instruction cache is not used optimally

Running `cargo-llvm-lines` on a small test program:
```rust
use moka::future::Cache;

pub struct Foo {
    cache: Cache<usize, usize>,
}

impl Foo {
    pub async fn bar(&self) -> usize {
        self.cache
            .entry(1)
            .or_insert_with_if(async { 1 }, |_| true)
            .await
            .into_value()
    }

    pub fn run(&self, runtime: tokio::runtime::Handle) -> usize {
        runtime.block_on(self.bar())
    }
}
```

gives me this these functions on the top:

```
  Lines                 Copies              Function name
  -----                 ------              -------------
  116847                3467                (TOTAL)
    6175 (5.3%,  5.3%)    57 (1.6%,  1.6%)  alloc::alloc::box_free
    3012 (2.6%,  7.9%)    21 (0.6%,  2.2%)  moka::cht::map::bucket::BucketArray<K,V>::probe_loop
    1988 (1.7%,  9.6%)     9 (0.3%,  2.5%)  moka::cht::map::bucket_array_ref::BucketArrayRef<K,V,S>::remove_entry_if_and
    1918 (1.6%, 11.2%)    48 (1.4%,  3.9%)  core::option::Option<T>::map
    1680 (1.4%, 12.6%)     5 (0.1%,  4.0%)  alloc::raw_vec::RawVec<T,A>::grow_amortized
    1417 (1.2%, 13.9%)    34 (1.0%,  5.0%)  core::result::Result<T,E>::map_err
    1215 (1.0%, 14.9%)     9 (0.3%,  5.3%)  moka::cht::map::bucket::BucketArray<K,V>::remove_if::{{closure}}
    1056 (0.9%, 15.8%)    18 (0.5%,  5.8%)  alloc::boxed::Box<T,A>::into_unique
    1008 (0.9%, 16.7%)     9 (0.3%,  6.1%)  <core::slice::iter::Iter<T> as core::iter::traits::iterator::Iterator>::next
[...]

PR #204, opened 2022-11-19
Deduplicate `do_get_with_hash`

Swatinem:
Due to the generic `read_recorder` closure, the `do_get_with_hash` function was generated twice in `cargo-llvm-lines` for a simple test.

This is the first improvement to #203 I have found

### DIFF
--- a/src/sync_base/base_cache.rs
+++ b/src/sync_base/base_cache.rs
@@ -294,13 +294,18 @@ where
                 } else {
                     // Valid entry.
                     let maybe_key = if need_key { Some(Arc::clone(k)) } else { None };
-                    Some((maybe_key, TrioArc::clone(entry), now))
+                    let v = entry.value.clone();
+                    let maybe_entry = if record_read {
+                        Some(TrioArc::clone(entry))
+                    } else {
+                        None
+                    };
+                    Some((maybe_key, v, maybe_entry, now))
                 }
             });
 
-        if let Some((maybe_key, entry, now)) = maybe_entry {
-            let v = entry.value.clone();
-            if record_read {
+        if let Some((maybe_key, v, maybe_entry, now)) = maybe_entry {
+            if let Some(entry) = maybe_entry {
                 self.record_read_op(ReadOp::Hit(hash, entry, now), now)
                     .expect("Failed to record a get op");
             }
--- a/src/sync_base/base_cache.rs
+++ b/src/sync_base/base_cache.rs
@@ -223,13 +223,8 @@ where
         K: Borrow<Q>,
         Q: Hash + Eq + ?Sized,
     {
-        // Define a closure to record a read op.
-        let record = |op, now| {
-            self.record_read_op(op, now)
-                .expect("Failed to record a get op");
-        };
         let ignore_if = None as Option<&mut fn(&V) -> bool>;
-        self.do_get_with_hash(key, hash, record, ignore_if, need_key)
+        self.do_get_with_hash(key, hash, true, ignore_if, need_key)
     }
 
     pub(crate) fn get_with_hash_but_ignore_if<Q, I>(
@@ -244,12 +239,7 @@ where
         Q: Hash + Eq + ?Sized,
         I: FnMut(&V) -> bool,
     {
-        // Define a closure to record a read op.
-        let record = |op, now| {
-            self.record_read_op(op, now)
-                .expect("Failed to record a get op");
-        };
-        self.do_get_with_hash(key, hash, record, ignore_if, need_key)
+        self.do_get_with_hash(key, hash, true, ignore_if, need_key)
     }
 
     pub(crate) fn get_with_hash_but_no_recording<Q, I>(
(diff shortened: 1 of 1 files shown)


tatsuya6502:
Thank you for tackling to this.

I actually wrote the original code by purpose; I was hoping `rustc` is smart enough to do some extra optimization for the duplicate function when it is given the closure that does nothing:

```rust
// Define a closure that skips to record a read op.
let record = |_op, _now| {};
```

Maybe I was right. I ran benchmark and found the change in this PR _slowed down the benchmark by 1.23%_ on my x86_64 Linux environment. (The number on the right bottom)

mokabench-dedup-2022-11-20

- Environment
    - CPU: Intel Core i7-12700F
    - OS: WSL2 Ubuntu 22.04 on Windows 11
    - Rust: 1.65.0 (897e37553 2022-11-02)

But I believe what you said on #203 is also true:

> This is suboptimal for a couple of reasons:
>
> * it generates multiple copied of functions that have to be compiled, hurting compile times.
> * may lead to binary bloat
> * may even hurt performance as the instruction cache is not used optimally

So maybe this is just one result and I may get opposite result with different benchmark on different environment?

tatsuya6502:
Here is how I ran the benchmark:

```console
$ git clone https://github.com/moka-rs/mokabench.git
$ cd mokabench

## Checkout the current head of `mini-moka` branch.
## (It will be merged into `master` soon)
$ git checkout acd6cde157c5f6e633a83d5a9514fb7bbc7b8617

## Follow [the README](https://github.com/moka-rs/mokabench/#download-the-trace-datasets)
## and place the ARC S3 trace dataset into `./datasets` directory.

## Apply the patch attached at the bottom of this comment.
$ git apply mokabench.patch

## Check what Moka version is used:
$ cargo tree -i moka

## Run the benchmark:
$ cargo run --release -- --insert-once

## Edit `Cargo.toml` to use the code for this PR, and run the benchmark again.

```

**mokabench.patch**

```diff
diff --git a/Cargo.toml b/Cargo.toml
index ecf3266..db73ffd 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,8 @@ optional = true
 # version = "0.10.0"
 git = "https://github.com/moka-rs/moka"
 branch = "master"
+# git = "https://github.com/Swatinem/moka"
+# branch = "do_get_with_hash"
 features = ["future"]
 
 [dependencies.moka09]
diff --git a/src/main.rs b/src/main.rs
index 2ec5981..ac4aae8 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -29,7 +29,8 @@ async fn main() -> anyhow::Result<()> {
 }
 
 async fn run_with_capacity(config: &Config, capacity: usize) -> anyhow::Result<()> {
-    const DEFAULT_NUM_CLIENTS_ARRAY: &[u16] = &[16, 24, 32, 40, 48];
+    // const DEFAULT_NUM_CLIENTS_ARRAY: &[u16] = &[16, 24, 32, 40, 48];
+    const DEFAULT_NUM_CLIENTS_ARRAY: &[u16] = &[4, 8, 16, 32];
 
     let num_clients_slice: &[u16] = if let Some(n) = &config.num_clients {
         &n
diff --git a/src/trace_file.rs b/src/trace_file.rs
index 638eefe..61f5fe7 100644
--- a/src/trace_file.rs
+++ b/src/trace_file.rs
@@ -61,7 +61,8 @@ impl TraceFile {
 
     pub fn default_capacities(&self) -> &[usize] {
         match self {
-            Self::S3 => &[100_000, 400_000, 800_000],
+            // Self::S3 => &[100_000, 400_000, 800_000],
+            Self::S3 => &[100_000, 800_000],
             Self::Ds1 => &[1_000_000, 4_000_000, 8_000_000],
             Self::Oltp => &[256, 512, 1_000, 2_000],
             Self::Loop => &[256, 512, 768, 1024],
```

tatsuya6502:
I wonder what the performance would be if we take your change, but change it further to avoid `TrioArc::clone` here when `record_read` is `false`:

https://github.com/moka-rs/moka/blob/bf5b57fa22fda6c1b7dfcbb000bb0855fd4cd630/src/sync_base/base_cache.rs#L297

Maybe one way to achieve this is to change the type of the variable `maybe_entry`

* `Option<(Option<Arc<K>>, TrioArc<ValueEntry<K, V>>)>`

to something like the followings?

* `Option<(Option<Arc<K>>, V, Option<TrioArc<ValueEntry<K, V>>>)>`

(I have to go now. I will try it later today or tomorrow. timezone:UTC+0800)

Swatinem:
Wow, that is a very thorough analysis.

I think I underestimated how well this crate is tuned for performance!

I applied your suggestion, but I doubt that will help. Now that I have understood the code paths a bit more, I believe its okay to have an extremely tuned/inlined version of the code for the ideal case of a cache hit, vs another cold copy that is only rarely hit.

Swatinem:
The recent work for #212 I believe has changed the structure of the code a lot, and also removed a bunch of functions that were creating tons of code.

I won’t be directly persuing this PR anymore, though I believe there might still be some compile-time and code-size gains to be had with some work.

PR #213, opened 2023-01-24
Fix the CI for Rust nightly with minimal dependency versions

tatsuya6502:
When testing minimal dependency versions, upgrade `openssl` to v0.10.39, otherwise building `openssl-sys` will fail on Linux.


tatsuya6502 (commented):
Merging.

PR #216, opened 2023-02-01
doc: Add examples for `build_with_hasher` method of cache builders

### LINKED ISSUE #214: how to use build_with_hash?
Hello! Thank you for this fantastic crate!!

Admittedly, I've never used an alternative hasher before so maybe I'm dong something wrong, but If I try with FxHasher, I get this:
```rust
error[E0308]: mismatched types
   --> foo/src/main.rs:28:28
    |
28  |         .build_with_hasher(FxBuildHasher::default())
    |          ----------------- ^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `RandomState`, found struct `BuildHasherDefault`
    |          |
    |          arguments to this function are incorrect
    |
    = note: expected struct `RandomState`
               found struct `BuildHasherDefault<FxHasher>`
```

So then i tried with ahash because, [in one of the issues](https://github.com/moka-rs/moka/issues/64) i found such an example
but then I get:
```rust
error[E0308]: mismatched types
    --> foo/src/main.rs:27:28
     |
27   |         .build_with_hasher(ahash::RandomState::default())
     |          ----------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `std::collections::hash_map::RandomState`, found struct `ahash::RandomState`
     |          |
     |          arguments to this function are incorrect
     |
```
Am I doing something wrong?

tatsuya6502:
This is a documentation update. It adds examples to cache builder's `build_with_hasher`. It will make it clear that user should specify the type of `S` explicitly. (`Cache<K, V, S>` instead of `Cache<K, V>`).

* * *
Fixes #214.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -49,7 +49,8 @@ linux_arm64_task:
   pin_deps_script: |
     if [ "v$RUST_VERSION" == "v1.51.0" ]; then
       echo 'Pinning some dependencies to specific versions'
-      rm Cargo.lock
+      rm -f Cargo.lock
+      sed -i 's/ahash = ".*"/ahash = "=0.7.6"/g' Cargo.toml
       cargo update -p dashmap --precise 5.2.0
       cargo update -p indexmap --precise 1.8.2
       cargo update -p hashbrown --precise 0.11.2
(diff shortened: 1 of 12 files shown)


tatsuya6502 (commented):
Merging.

Issue #212, opened 2023-01-16
Moka `get_with` inflates future size by ~7x


Swatinem:
Consider the following example:

```rust
fn main() {
    let cache = moka::future::Cache::new(1);
    let get_fut = async {
        let buf = [0u8; 1024];
        async {}.await;
        drop(buf);
    };
    println!("get_fut size: {}", std::mem::size_of_val(&get_fut));
    let moka_fut = cache.get_with((), get_fut);
    println!("moka_fut size: {}", std::mem::size_of_val(&moka_fut));
}
```

This outputs the following (even in `--release` mode):

```
get_fut size: 1026
moka_fut size: 7688
```

The resulting future grows by pretty much 7x the size of the future I provide.

The problem does not seem to be affected by https://github.com/rust-lang/rust/issues/62321 though.
If I comment out the two `println` statements, I can use a nightly compiler to log the type sizes, using `cargo +nightly rustc -- -Zprint-type-sizes`:

<details>
<summary>-Zprint-type-sizes</summary>

```
print-type-size type: `[async fn body@moka::future::Cache<(), ()>::get_with<[async block@src\main.rs:3:19: 7:6]>::{closure#0}]`: 7688 bytes, alignment: 8 bytes
print-type-size     discriminant: 1 bytes
print-type-size     variant `Suspend0`: 7684 bytes
print-type-size         field `.key`: 0 bytes, offset: 0 bytes, alignment: 1 bytes
print-type-size         field `.key`: 0 bytes
print-type-size         field `.__awaitee`: 6648 bytes
print-type-size         field `.self`: 8 bytes
print-type-size         field `.init`: 1026 bytes
print-type-size         field `..generator_field2`: 1 bytes
print-type-size         field `..generator_field3`: 1 bytes
print-type-size     variant `Unresumed`: 7682 bytes
print-type-size         field `.key`: 0 bytes, offset: 0 bytes, alignment: 1 bytes
print-type-size         padding: 6648 bytes
print-type-size         field `.self`: 8 bytes, alignment: 8 bytes
print-type-size         field `.init`: 1026 bytes
print-type-size     variant `Returned`: 7682 bytes
print-type-size         field `.key`: 0 bytes, offset: 0 bytes, alignment: 1 bytes
print-type-size         padding: 6648 bytes
print-type-size         field `.self`: 8 bytes, alignment: 8 bytes
print-type-size         field `.init`: 1026 bytes
print-type-size     variant `Panicked`: 7682 bytes
[...]

PR #220, opened 2023-02-04
Reduce the size of the future returned by async `get_with` and friend methods (v0.10)

tatsuya6502:
This PR will mitigate #212 for upcoming Moka v0.10.0 release.

## Changes

This PR does the same thing to PR #222, but on the `master` branch instead of the `maint-09` branch.

`future::Cache`:

1. To mitigate current `rustc`'s optimization issue, switched the internally using type for the `init` future from `impl Future` to a reference `Pin<&mut impl Future>`.
2. Eliminated an internal async function `ValueInitializer:init_or_read(..., init, ...)` from the call path.

## Updated Methods:

`future::Cache`:

- [x] `get_with(...)`
- [x] `get_with_if(...)`
- [x] `optionally_get_with(...)`
- [x] `try_get_with(...)`
- [x] `get_with_by_ref(...)`
- [x] `optionally_get_with_by_ref(...)`
- [x] `try_get_with_by_ref(...)`
- [x] `entry(...).or_insert_with(...)`
- [x] `entry(...).or_insert_with_if(...)`
- [x] `entry(...).or_optionally_insert_with(...)`
- [x] `entry(...).or_try_insert_with(...)`
- [x] `entry_by_ref(...).or_insert_with(...)`
- [x] `entry_by_ref(...).or_insert_with_if(...)`
- [x] `entry_by_ref(...).or_optionally_insert_with(...)`
- [x] `entry_by_ref(...).or_try_insert_with(...)`

### DIFF
--- a/src/future/cache.rs
+++ b/src/future/cache.rs
@@ -27,13 +27,14 @@ use std::{
     fmt,
     future::Future,
     hash::{BuildHasher, Hash},
+    pin::Pin,
     sync::Arc,
     time::Duration,
 };
 
 //
-// Macros to mitigate the compiler issue that inflates the size of the `init` future
-// in `get_with` and friends methods:
+// Macros to mitigate the compiler issue that inflates the size of future returned
+// from `get_with` and friends methods:
 //
 // - https://github.com/moka-rs/moka/issues/212
 // - https://swatinem.de/blog/future-size/
@@ -43,10 +44,9 @@ use std::{
 // benefits of reducing the future size outweigh this drawback.
 //
 macro_rules! insert_with_hash_and_fun_m(
-    ( $self:ident, $key:expr, $hash:expr, $get:expr, $init:expr, $replace_if:expr, $need_key:expr ) => {
+    ( $self:ident, $key:expr, $hash:expr, $get:expr, $init:ident, $replace_if:expr, $need_key:expr ) => {
         {
             use futures_util::FutureExt;
-
 
             let insert = |v| $self.insert_with_hash($key.clone(), $hash, v).boxed();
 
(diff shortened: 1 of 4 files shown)


src/future/value_initializer.rs
```diff
@@ -345,6 +237,42 @@ where
...
+    }
+
+    /// Returns the `type_id` for `get_with` method of cache.
+    pub(crate) fn type_id_for_get_with() -> TypeId {
+        TypeId::of::<()>()
+    }
+
+    /// Returns the `type_id` for `optionally_get_with` method of cache.
+    pub(crate) fn type_id_for_optionally_get_with() -> TypeId {
+        TypeId::of::<()>()
```
tatsuya6502, line 269:
```suggestion
        TypeId::of::<OptionallyNone>()
```

Swatinem:
Hm, The usage of macros is very interesting, I haven’t thought of that one. Though admittedly it is a bit sad that futures have such an overhead that one workaround is to replace them with something as unergonomic as macros. :-D

src/future/cache.rs
```diff
@@ -948,9 +950,11 @@ where
         init: impl Future<Output = V>,
         replace_if: impl FnMut(&V) -> bool,
     ) -> V {
+        futures_util::pin_mut!(init);
         let hash = self.base.hash(&key);
         let key = Arc::new(key);
-        self.get_or_insert_with_hash_and_fun(key, hash, init, Some(replace_if), false)
+        let replace_if = Some(replace_if);
+        self.get_or_insert_with_hash_and_fun(key, hash, init, replace_if, false)
```
tatsuya6502, line 957: Undo this change:

```suggestion
        self.get_or_insert_with_hash_and_fun(key, hash, init, Some(replace_if), false)
```

tatsuya6502 (commented):
I will merge this into the `master` branch once all CI jobs are completed. I will do a squash merge.

PR #224, opened 2023-02-07
Refactor an internal module `sync::value_initializer`

tatsuya6502:
Refactor an internal module `sync::value_initializer` to match the new structure of `future::value_initializer` introduced by #220.

- Eliminated an internal method `ValueInitializer:init_or_read(..., init, ...)` from the call path.

Also remove unnecessary `Arc::clone`s from both `sync::Cache` and `future::Cache`.

### DIFF
--- a/src/future/cache.rs
+++ b/src/future/cache.rs
@@ -1501,7 +1501,7 @@ where
 
         match self
             .value_initializer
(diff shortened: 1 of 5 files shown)


tatsuya6502 (commented):
Merging.

PR #232, opened 2023-02-18
Update the feature comparison table in the README

tatsuya6502:
- Update the info about `quick_cache` to `v0.2`.
- Add a comment to `moka` that the background threads will be removed from `v0.11`.

CC: @arthurprs


tatsuya6502 (commented):
Merging.

PR #238, opened 2023-03-04
Do not cache anything when max capacity is zero

### LINKED ISSUE #230: Cache with max capacity of zero should not cache anything
## Description

All Moka versions (up to v0.10.0) are affected by this issue.

When a cache is configured with the max capacity of zero, users would expect that it should not cache anything. However, the current cache implementations are to cache everything for a short period of time (up to 0.5 seconds). This is because these implementations temporary admit new entries no matter if the cache is full or not, and evict the entries later in batch.

This behavior is not intuitive, so it needs to be changed to not caching anything.

```rust
// Cargo.toml
// [dependencies]
// moka = { version = "0.10.0", features = ["future"] }
// tokio = { version = "1.25.0", features = ["rt-multi-thread", "time"] }

fn main() {
    sync_cache();

    let rt = tokio::runtime::Runtime::new().unwrap();
    rt.block_on(future_cache());
}

fn sync_cache() {
    let cache = moka::sync::Cache::new(0);
    cache.insert(0, ());

    dbg!(cache.contains_key(&0));
    std::thread::sleep(std::time::Duration::from_millis(550));
    dbg!(cache.contains_key(&0));
}

async fn future_cache() {
    let cache = moka::future::Cache::new(0);
    cache.insert(1, ()).await;

    dbg!(cache.contains_key(&1));
    tokio::time::sleep(std::time::Duration::from_millis(550)).await;
    dbg!(cache.contains_key(&1));
}
```

```console
$ cargo run
[src/main.rs:12] cache.contains_key(&0) = true
[src/main.rs:14] cache.contains_key(&0) = false
[src/main.rs:21] cache.contains_key(&1) = true
[src/main.rs:23] cache.contains_key(&1) = false
```

## Workaround

Users can workaround this issue by setting `time_to_live` to `ZERO`. This will still cache new entries for a short period of time, but `get*` methods will never return the entries as they are already expired.

```rust
        .max_capacity(0)
        .time_to_live(std::time::Duration::ZERO)
        .build();
```

CC: @Swatinem

tatsuya6502:
To fix #230, make the cache implementations to store nothing when the max capacity is set to 0.

## Changes

When the max capacity is set to 0, cache will do the followings:

- Create the internal concurrent hash map with initial capacity of 0.
- Do not enable the frequency sketch.
- Make the following methods to do nothing and return immediately:
    - internal `insert_with_hash`
    - internal `do_get_with_hash`
    - external `sync`

Also added unit test called `max_capacity_zero`.

## Verification

- [x] Ran all unit tests and they all passed.
    - command: `RUSTFLAGS='--cfg skeptic --cfg trybuild' cargo test -F future`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Moka Cache &mdash; Change Log
 
+## Version 0.10.1
+
+### Changed
+
+- Now `sync` and `future` caches will not cache anything when the max capacity is set
+  to zero ([#230][gh-issue-0230]):
+    - Previously, they would cache some entries for short time (< 0.5 secs) even
+      though the max capacity is zero.
+
+
 ## Version 0.10.0
 
 ### Breaking Changes
@@ -575,6 +585,7 @@ The minimum supported Rust version (MSRV) is now 1.51.0 (2021-03-25).
 [gh-Swatinem]: https://github.com/Swatinem
 [gh-tinou98]: https://github.com/tinou98
 
+[gh-issue-0230]: https://github.com/moka-rs/moka/issues/230/
 [gh-issue-0212]: https://github.com/moka-rs/moka/issues/212/
 [gh-issue-0207]: https://github.com/moka-rs/moka/issues/207/
 [gh-issue-0162]: https://github.com/moka-rs/moka/issues/162/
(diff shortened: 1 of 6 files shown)


tatsuya6502 (commented):
Merging.

PR #241, opened 2023-03-11
Avoid a deprecation warning from scheduled-thread-pool v0.2.7

tatsuya6502:
- Replace the deprecated `ScheduledThreadPool::with_name` method with the recommended `builder` method.
- Since the `builder` method was just added to v0.2.7, had to change the minimum version of scheduled-thread-pool from v0.2.6 to v0.2.7.


tatsuya6502 (commented):
Merging.

PR #244, opened 2023-03-24
Fix the CI for MSRV and Rust nightly (cc, thiserror, serde)

tatsuya6502:
- When testing minimal dependency versions, upgrade `cc` to v1.0.61, otherwise building `openssl-sys` will fail on Linux.
- When testing MSRV 1.51.0, downgrade `thiserror` and `serde` to v1.0.39 and v1.0.156
respectively.


tatsuya6502 (commented):
Merging.

PR #246, opened 2023-03-25
`future::Cache`'s `.blocking().invalidate(key)` should trigger the eviction listener

### LINKED ISSUE #242: `.blocking().invalidate(key)` doesn't trigger eviction listener
I've spotted that a `moka::future::Cache` doesn't fire its eviction listener if an entry is manually invalidated with the blocking op, but I can fire the eviction listener if using `.invalidate_all()` or the sync cache. The below examples should repro this behaviour. 

```
#[tokio::test]
async fn invalidate_all_triggers_listener() {
    let cache = moka::future::Cache::builder()
        .eviction_listener_with_queued_delivery_mode(|k, v, c| {
            println!("evicted key: {}, value: {}, cause: {:?}", k, v, c);
        })
        .build();
    cache.get_with("foo", async { "all" }).await;
    cache.invalidate_all();
    tokio::time::sleep(Duration::from_secs(1)).await;
}

#[tokio::test]
async fn sync_triggers_listener() {
        .eviction_listener(|k, v, c| {
            println!("evicted key: {}, value: {}, cause: {:?}", k, v, c);
        })
        .build();
    cache.insert("foo", "sync");
    cache.invalidate("foo");
}

#[tokio::test]
async fn blocking_invalidate_should_trigger_listener() {
    let cache = moka::future::Cache::builder()
        .eviction_listener_with_queued_delivery_mode(|k, v, c| {
            println!("evicted key: {}, value: {}, cause: {:?}", k, v, c);
        })
        .build();
    cache.get_with("foo", async { "blocking" }).await;
    cache.blocking().invalidate("foo");
    tokio::time::sleep(Duration::from_secs(1)).await;
}
```

For these tests the eviction handler only fires for the first two:
```
running 3 tests
evicted key: foo, value: sync, cause: Explicit
test sync_triggers_listener ... ok
evicted key: foo, value: all, cause: Explicit
test invalidate_all_triggers_listener ... ok
test blocking_invalidate_should_trigger_listener ... ok
```

My understanding of invalidate is that it is a manually triggered cache eviction, is that correct? If so I think this may be a bug.

tatsuya6502:
Fixes #242.

## Changes

- Fix an internal `do_blocking_invalidate` method to notify the eviction listener when an entry is evicted.
- Update a test to verify that the eviction listener is notified when an entry is evicted by `.blocking().invalidate(key)`.

## Verification Methods

- [x] Verify the updated test `basic_single_blocking_api` passes.
- [x] Verify the reproducing code in #242 no longer reproduces the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Version 0.10.1
 
+### Fixed
+
+- Fixed a bug that `future` cache's `blocking().invalidate(key)` method does not
+  trigger the eviction listener. ([#242][gh-issue-0242])
+
 ### Changed
 
 - Now `sync` and `future` caches will not cache anything when the max capacity is set
@@ -585,6 +590,7 @@ The minimum supported Rust version (MSRV) is now 1.51.0 (2021-03-25).
 [gh-Swatinem]: https://github.com/Swatinem
 [gh-tinou98]: https://github.com/tinou98
 
+[gh-issue-0242]: https://github.com/moka-rs/moka/issues/242/
 [gh-issue-0230]: https://github.com/moka-rs/moka/issues/230/
 [gh-issue-0212]: https://github.com/moka-rs/moka/issues/212/
 [gh-issue-0207]: https://github.com/moka-rs/moka/issues/207/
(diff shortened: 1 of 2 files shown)


tatsuya6502 (commented):
Merging.

Issue #227, opened 2023-02-10
Entry `and_compute` method for performing an insert, update, or remove operation with single closure


tatsuya6502:
The entry API was added to Moka v0.10.0. It has some methods such as `entry_by_ref(&key).or_insert_with(|| ...)` to atomically insert a value when key does not exist. This issue will add another method `and_compute` to the entry API for more complex tasks. It is similar to the [`compute` method] in Java Caffeine cache. Moka user can use `and_compute` to insert, update, or remove a value for a key with single closure or function.

[`compute` method]: https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/latest/com.github.benmanes.caffeine/com/github/benmanes/caffeine/cache/Policy.VarExpiration.html#compute(K,java.util.function.BiFunction,java.time.Duration)

**A new entry API: `and_compute` method**

```rust
use moka::{Entry, ops::{Op, PerformedOp}};

// For moka::sync::Cache's entry API
// `K` and `V` are the key and value type of the cache.
pub fn and_compute<F>(self, f: F) -> (Option<Entry<K, V>>, PerformedOp)
where
    F: FnOnce(Option<&V> -> Op<V>);

// For moka::future::Cache's entry API
pub async fn and_compute<'a, F, Fut>(self, f: F) -> (Option<Entry<K, V>>, PerformedOp)
where
    F: FnOnce(Option<&'a V>) -> Fut,
    Fut: Future<Output = Op<V>> + 'a;
```

`moka::ops::Op` and `moka::ops::PerformedOp` will be defined as:

```rust
/// Instructs the `and_compute` method how to modify the cache entry.
pub enum Op<V> {
    /// No-op. Do not modify the cache entry.
    Nop,
    /// Insert or replace the value of the cache entry.
    Put(V),
    /// Invalidate the cache entry.
    Invalidate,
}

/// Will be returned from `and_compute` method to indicate what kind of
/// operation was performed.
pub enum PerformedOp {
    /// The entry did not exist, or already existed but was not modified.
    Nop,
    /// The entry did not exist and inserted.
    Inserted,
    /// The entry already existed and its value was updated.
    Updated,
    /// The entry existed and was invalidated.
    /// Note: If `and_compute` tried to invalidate a not-exiting entry, 
    /// `Nop` will be returned instead of `Invalidated`
    Invalidated,
}
```

## Example

Here is an example of a future cache holding counters. (from a question in #179)

```rust
[...]

PR #248, opened 2023-03-31
Per-entry expiration

### LINKED ISSUE #231: Proposal: support insert key with expire time?
Now moka has a global expire time for all keys, can we insert key with expire time to make some keys different?

tatsuya6502:
Fixes #231.

## Action Items

1. [x] Implement the per-key expiration:
    * [x] a. Implement a new data structure called hierarchical timing wheels ([pdf](http://www.cs.columbia.edu/~nahum/w6998/papers/ton97-timing-wheels.pdf)) to manage the per-key expiration. (e.g. Caffeine's [TimerWheel.java](https://github.com/ben-manes/caffeine/blob/1e719aaa989dcea4356f593095d9c8cb43995114/caffeine/src/main/java/com/github/benmanes/caffeine/cache/TimerWheel.java))
    * [x] b. Add test cases for the new data structure.
    * [x] c. Update the internal functions of `sync` and `future` caches to use the new data structure for:
        * [x] cache write operations such as `insert` and `invalidate`.
        * [x] finding and evicting expired entries.
    * ~~d. Update the `sync` and `future` cache API to provide `expires_after` method.~~ ([#227 comment](https://github.com/moka-rs/moka/issues/227#issuecomment-1435800906))
    * [x] d. Update the `sync` and `future` caches to be configured with `expire_after` at the cache creation time. `expire_after` takes an expiry that implements `Expiry` trait. ([#227 comment](https://github.com/moka-rs/moka/issues/227#issuecomment-1436141367))
    * [x] e. Add test cases for the per-key expiration.
    * [x] f. To save some runtime memory, refactor existing per-cache TTL and Time-to-Idle codes to utilize the struct fields added for per-key expiration.
        * (Remove the key and hash fields from the deque node structs, and utilize the key and hash fields added to the entry-info struct)
2. [ ] Write the documentation for the per-key expiration with sample codes.
3. [x] Update a benchmarking tool called Mokabench. (Currently on [per-key-expiration](https://github.com/moka-rs/mokabench/tree/per-key-expiration) branch)


ben-manes:
fwiw, mine is pretty low-level as I wanted to minimize time-space overhead, and could couple it to the implementation. It is a little harder to think through since I wanted to avoid an expensive division / modulus instruction in favor of bit manipulations and didn't want to rely on the compiler as I cannot assert its optimizations in Java. I couldn't find a good reference to crib from, though I did ask https://github.com/tokio-rs/tokio-timer/issues/30 about their hashed wheel. I guess that inspired them since they [rewrote it](https://tokio.rs/blog/2018-03-timers) and theirs is more generalized, so it might be another reference or easy to borrow directly.

tatsuya6502:
@Millione @Swatinem 

Hi. The per-entry expiration feature is ready to play with. It has not been thoroughly tested yet, and is definitely not ready for a release yet, but I would like to get some feedback from you guys.

**Cargo.toml**

```toml
[dependencies]
moka = { git = "https://github.com/moka-rs/moka.git", branch = "per-entry-expiration", features = ["future"] }
```

As for the API, I took [the suggestion](https://github.com/moka-rs/moka/issues/227#issuecomment-1436141367) from @ben-manes, the author of [the Caffeine cache](https://github.com/ben-manes/caffeine), to use a callback evaluator `expiry` to determine the expiration time of an entry, rather than directly supplying the expiration time by `get_with` or `insert`, etc. By doing so, we can support more flexible expiration policies; not only time-to-live, but also time-to-idle.

The `Expiry` trait provides three methods with default implementations:

```rust
pub trait Expiry<K, V> {
    /// Specifies that the entry should be automatically removed from the cache once
    /// the duration has elapsed after the entry's creation. Returning `None`
    /// indicates no expiration for the entry.
    /// ...
    fn expire_after_create(&self, key: &K, value: &V, current_time: Instant) -> Option<Duration> {
        None  // No expiry.
    }

    /// Specifies that the entry should be automatically removed from the cache once
    /// the duration has elapsed after its last read. Returning `None` indicates no
    /// expiration for the entry. Returning `current_duration` will not modify the
    /// expiration time.
    /// ...
    fn expire_after_read(
        &self,
        key: &K,
        value: &V,
        current_time: Instant,
        // Represents the duration until the entry expires.
        current_duration: Option<Duration>,
    ) -> Option<Duration> {
        current_duration  // No change.
    }

    /// Specifies that the entry should be automatically removed from the cache once
    /// the duration has elapsed after the replacement of its value. Returning `None`
    /// indicates no expiration for the entry. Returning `current_duration` will not
    /// modify the expiration time.
    /// ...
    fn expire_after_update(
        &self,
        key: &K,
        value: &V,
        current_time: Instant,
        // Represents the duration until the entry expires.
        current_duration: Option<Duration>,
    ) -> Option<Duration> {
        current_duration  // No change.
    }
}
```

For more details about the methods, see the `Expiry` trait in [the `policy` module](https://github.com/moka-rs/moka/blob/per-entry-expiration/src/policy.rs).

Here is a working example. You will create a type that implements the `Expiry` trait, and pass it to the `expire_after` method of a cache builder to set it to a cache.

```rust
// Cargo.toml
// [dependencies]
// moka = { git = "https://github.com/moka-rs/moka.git", branch = "per-entry-expiration", features = ["future"] }
// tokio = { version = "1.27.0", features = ["rt-multi-thread", "time", "macros"] }

use moka::{future::Cache, Expiry};
use std::time::{Duration, Instant};

// In this example, we will create a `future::Cache` with `u32` as the key, and
// `(Expiration, String)` as the value. `Expiration` is an enum to represent the
// expiration of the value, and `String` is the application data of the value.

/// An enum to represent the expiration of a value.
#[derive(Clone, Copy, Debug, Eq, PartialEq)]
pub enum Expiration {
    /// The value never expires.
    Never,
    /// The value expires after a short time. (5 seconds in this example)
    AfterShortTime,
    /// The value expires after a long time. (15 seconds in this example)
    AfterLongTime,
}

impl Expiration {
    /// Returns the duration of this expiration.
    pub fn as_duration(&self) -> Option<Duration> {
        match self {
            Expiration::Never => None,
            Expiration::AfterShortTime => Some(Duration::from_secs(5)),
            Expiration::AfterLongTime => Some(Duration::from_secs(15)),
        }
    }
}

/// An expiry that implements `moka::Expiry` trait. `Expiry` trait provides the
/// default implementations of three callback methods `expire_after_create`,
/// `expire_after_read`, and `expire_after_update`.
///
/// In this example, we only override the `expire_after_create` method.
pub struct MyExpiry;

impl Expiry<u32, (Expiration, String)> for MyExpiry {
    /// Returns the duration of the expiration of the value that was just
    /// created.
    fn expire_after_create(
        &self,
        _key: &u32,
        value: &(Expiration, String),
        _current_time: Instant,
    ) -> Option<Duration> {
        let duration = value.0.as_duration();
        println!("MyExpiry: expire_after_create called with key {_key} and value {value:?}. Returning {duration:?}.");
        duration
    }
}

#[tokio::main]
async fn main() {
    // Create a `Cache<u32, (Expiration, String)>` with an expiry `MyExpiry` and
    // eviction listener.
    let expiry = MyExpiry;

    let eviction_listener = |key, _value, cause| {
        println!("Evicted key {key}. Cause: {cause:?}");
    };

    let cache = Cache::builder()
        .max_capacity(100)
        .expire_after(expiry)
        .eviction_listener_with_queued_delivery_mode(eviction_listener)
        .build();

    // Insert some entries into the cache with different expirations.
    cache
        .get_with(0, async { (Expiration::AfterShortTime, "a".to_string()) })
        .await;

    cache
        .get_with(1, async { (Expiration::AfterLongTime, "b".to_string()) })
        .await;

    cache
        .get_with(2, async { (Expiration::Never, "c".to_string()) })
        .await;

    // Verify that all the inserted entries exist.
    assert!(cache.contains_key(&0));
    assert!(cache.contains_key(&1));
    assert!(cache.contains_key(&2));

    // Sleep for 6 seconds. Key 0 should be expired.
    println!("\nSleeping for 6 seconds...\n");
    tokio::time::sleep(Duration::from_secs(6)).await;
    println!("Entry count: {}", cache.entry_count());

    // Verify that key 0 has been evicted.
    assert!(!cache.contains_key(&0));
    assert!(cache.contains_key(&1));
    assert!(cache.contains_key(&2));

    // Sleep for 10 more seconds. Key 1 should be expired.
    println!("\nSleeping for 10 seconds...\n");
    tokio::time::sleep(Duration::from_secs(10)).await;
    println!("Entry count: {}", cache.entry_count());

    // Verify that key 1 has been evicted.
    assert!(!cache.contains_key(&1));
    assert!(cache.contains_key(&2));

    // Manually invalidate key 2.
    cache.invalidate(&2).await;
    assert!(!cache.contains_key(&2));

    println!("\nSleeping for a second...\n");
    tokio::time::sleep(Duration::from_secs(1)).await;
    println!("Entry count: {}", cache.entry_count());

    println!("\nDone!");
}
```

**Outputs**

```console
MyExpiry: expire_after_create called with key 0 and value (AfterShortTime, "a"). Returning Some(5s).
MyExpiry: expire_after_create called with key 1 and value (AfterLongTime, "b"). Returning Some(15s).
MyExpiry: expire_after_create called with key 2 and value (Never, "c"). Returning None.

Sleeping for 6 seconds...

Evicted key 0. Cause: Expired
Entry count: 2

Sleeping for 10 seconds...

Evicted key 1. Cause: Expired
Entry count: 1

Sleeping for a second...

Evicted key 2. Cause: Explicit
Entry count: 0

Done!
```

tatsuya6502:
Just a heads-up. I added an argument `last_modified_at: Instant` to `Expiry::expire_after_read` method (commit https://github.com/moka-rs/moka/pull/248/commits/dc5fd49206f5ff61f8e5ec8e17fde38d526cf7f5). I updated the example in the previous comment.

tatsuya6502:
I found that the current implementation (commit https://github.com/moka-rs/moka/commit/dc5fd49206f5ff61f8e5ec8e17fde38d526cf7f5) has a bug in per-entry expiration relating to  `invalidate(&key)` method. It seems race conditions (?) cause internal timer node to be dropped more than once, causing panic and/or segmentation fault.

Until fixed, please do not use per-entry expiration if you also use `invalidate` method.

**[Mokabench](https://github.com/moka-rs/mokabench) Results**

```
$ ./target/release/mokabench --num-clients 16,32 --ttl 3 --tti 1 --per-key-expiration --invalidate
...

Cache, Max Capacity, Clients, Inserts, Reads, Hit Ratio, Duration Secs

'./target/release/mokabench --nu…' terminated by signal SIGSEGV (Address boundary error)
```

```
$ ./target/release/mokabench --num-clients 16,32 --ttl 3 --tti 1 --per-key-expiration --invalidate
...

Cache, Max Capacity, Clients, Inserts, Reads, Hit Ratio, Duration Secs

thread 'moka-housekeeper-4' panicked at 'index out of bounds: the len is 5 but the index is 128', ... /moka/src/common/timer_wheel.rs:243:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'moka-housekeeper-7' panicked at 'index out of bounds: the len is 5 but the index is 200', ... /moka/src/common/timer_wheel.rs:243:17
^C
```

Millione:
> Until fixed, please do not use per-entry expiration if you also use `invalidate` method.

I tried the per-entry expiration without using the `invalidate` method, but it also seems to have something wrong with the implementation, cause hit rate will increase for a while and then suddenly hang without any further information.

tatsuya6502:
@Millione

Thank you for trying it! 

> but it also seems to have something wrong with the implementation, cause hit rate will increase for a while and then suddenly hang without any further information.

Sorry for the problem. It sounds like a dead-lock or infinite loop happening. 

I will try to reproduce it later today, but it could be related with your `Expiry` implementation. So, if you can get more information, that would be very helpful.

Firstly, check if an infinite loop is happening in Moka's housekeeper thread. Run `top` or something and see if one of the CPU cores has ~100% utilization.

Secondly, check what your application is doing by taking stack traces of all threads in your application. This can be done by using `rust-lldb` or `rust-gdb`:

**rust-lldb**

```console
## You need Rust toolchain and `lldb` installed.

$ rust-lldb
(lldb) process attach --pid <PID>
(lldb) bt all
```

**rust-gdb**

```console
## You need Rust toolchain and `gdb` installed.

$ rust-gdb
(gdb) attach <PID>
(gdb) thread apply all bt
```

If an infinite loop is happening, look for "moka-housekeeper-*" threads. Maybe one of them is executing a function in `TimerWheel` or `base_cache::Inner`.

If a deadlock is happening, look for a "moka-housekeeper-*" thread waiting for a lock in a function in `TimerWheel` or `base_cache::Inner`, and also look for a user thread (maybe "tokio-thread-pool-*") waiting/executing in a function in your `Expiry` implementation.

You can also paste the full stack traces to this issue if you want. Thanks!

src/common/timer_wheel.rs
```diff
@@ -237,11 +260,13 @@ impl<K> TimerWheel<K> {
...
-            let index = index.load(Ordering::Acquire);
-            if level != u8::MAX && index != u8::MAX {
-                self.wheels[level as usize][index as usize].unlink(node);
+            let lev = level.load(Ordering::Acquire);
+            let idx = index.load(Ordering::Acquire);
+            if lev != u8::MAX && idx != u8::MAX {
+                self.wheels[lev as usize][idx as usize].unlink(node);
             }
+            level.store(u8::MAX, Ordering::Release);
+            index.store(u8::MAX, Ordering::Release);
```
Swatinem: I’m absolutely no expert on atomics, etc, but I’m curious here:
- Why is this two separate `AtomicU8` values, instead of a single `AtomicU16` with two masked values?
- As these are not cacheline-aligned, wouldn’t these two separate values clash directly? Probably not a big deal, but as multiple threads can probably interleave here, two load + two store operations per thread have a higher likelyhood to slow things down, no?
- I believe there is still a race condition between the `load` and, `unlink` and `store`. Two threads can first load the values and execute the `unlink`, before either thread sets that to `u8::MAX`
- Wouldn’t it be better to use an atomic exchange to change this to `u8::MAX` (or `u16::MAX` if you merge the two flags) before checking the flag, so that any competing thread will never execute `unlink` twice?

tatsuya6502:
@Millione

I could not reproduce the hung issue. I used the head of `per-entry-expiration` branch containing the fix for the `invalidate` bug. But I do not think that fix would also solve the hung issue. So I think it will be still there.

Please check if the hung issue remains (when you have time). If so, any additional information would be helpful.

src/common/timer_wheel.rs
tatsuya6502: Thank you for reviewing!

As for using two separate `AtomicU8` values, there is not much reason; I just wanted two integers that are `Send` + `Sync`. As you said, using a single `AtomicU16` should be more efficient (and atomic) so I will change them so.

As for the race condition, we do not have to be bothered in `TimerWheel` code because it is guaranteed that no more than one thread will call a method of single `TimerWheel` instance at the same time. About half the code in `base_cache` module are executed by the `sync` method of `Inner` type. And `sync` method is guarded by a cache-wide lock `self.deques.lock()`. This is by design; `sync` applies cache reads/writes logs in batch to reduce the lock contentions.

Other data structures including `Deque`s and `FrequencySketch` are guarded by the same cache-wide lock.

Maybe I should add source code comments and/or divide the module into smaller to make this clear.

Swatinem:
I agree that a callback-based API makes sense to have maximum flexibility and control. However it does introduce a bit of boilerplate, as I have to still keep an `Instant` around in my value. We calculate that TTL in our `init` future directly.
However that boilerplate is not quite as bad: https://github.com/getsentry/symbolicator/pull/1136/files

I have some infrastructure issues putting this into a canary deployment anyway, so I guess that was actually a good thing to have delayed me, given this seems to still be deadlocking.

Millione:
@tatsuya6502 

Sorry for the late reply, I've been quite busy for a few days. I will check it again with the latest branch when I have time and get back to you with the result and more information as soon as possible.

Also, the expiry implementation looks like this. Because it's an internal project running in the product environment, it's a bit harder to test and unfortunately I can't provide you the full backtrace information.
```rust
pub struct MyExpiry;

pub struct MyResp {
  ttl: std::time::Duration,
}

impl Expiry<faststr::FastStr, Option<Arc<RwLock<MyResp>>>> for MyExpiry {
    fn expire_after_create(
        &self,
        _key: &faststr::FastStr,
        value: &Option<Arc<RwLock<MyResp>>>,
        _current_time: std::time::Instant,
    ) -> Option<Duration> {
        if let Some(resp) = value {
            Some(resp.read().ttl)
        } else {
            Some(std::time::Duration::from_secs(60 * 60))
        }
    }
}
```

tatsuya6502:
@Swatinem @Millione

Thank you for the updates and feedback. Millione, please take your time. And sorry about the quality.

I have been running very intense, multi-threaded workloads using Mokabench for few hours, and things are stable now (no panics, no write stalls, etc.). My expiry implementation is [here](https://github.com/moka-rs/mokabench/blob/6143ba953a16bb0b4c3671dfb13982ff351c0b9c/src/cache/moka_driver.rs#L52-L120).

I hope using the latest branch may help. Here are last four commits with fixes to prevent panics and an infinite loop.

recent-fixes

- The first highlight fixed an infinite loop in `TimerWheel`.
    - When it happens, cache writes will eventually stall as internal bounded channel will become full.
- The second highlight removed some careless `unwrap()`s on duration calculations.
    - It may cause panic if an expiry method returns a very short duration (such as zero).
    - When the panic happens, cache writes will eventually stall too.
- The last highlight was the bug related to `invalidate(&key)` method.
    - When it happens, the process can be aborted by segmentation fault.

On next few days, I will check if there are any memory leaks in unsafe codes. And also write more unit tests.

Millione:
@tatsuya6502 

Thank you very much for the great work you have done and for providing us with detailed information on how to solve the problem. It definitely helps a lot. Much appreciated.

src/common/timer_wheel.rs
tatsuya6502: It turned out I did not need interior mutability for them 😅 because the `TimerWheel`, the owner of `TimerNode`s, is guarded by a mutex. I replaced these two `AtomicU8`s with a single `Option<(u8, u8)>`.

I am running some load tests now. I will check-in the change once I get some confidence.

tatsuya6502: > I replaced these two `AtomicU8`s with a single `Option<(u8, u8)>`.

OK. I checked-in the change: [`ac4d32d`#diff-L86-R86](https://github.com/moka-rs/moka/pull/248/commits/ac4d32de4ef1deb925c50c227cbccec957effcf4#diff-a9ca7964752b28bb55848a11e88855234c65c66430b6fc82f2dacaf265c55fc8L86-R86)

Millione:
> I tried the per-entry expiration without using the `invalidate` method, but it also seems to have something wrong with the implementation, cause hit rate will increase for a while and then suddenly hang without any further information.

@tatsuya6502 Well, I have refactored my code and can no longer reproduce it. So I think it might be my own code problem.

tatsuya6502:
@Millione — Thanks a lot for testing it again! That is a good news.

tatsuya6502:
> On next few days, I will check if there are any memory leaks in unsafe codes.

I checked it few days ago and looked okay.

> And also write more unit tests.

I did not have chance to do it, and could not write doc with examples.

I believe this PR got all the planned functionalities already. I am going to review and merge this into the `master` branch. I will follow up with another PR with more unit tests and documentation.

src/sync_base/base_cache.rs
```diff
@@ -2037,6 +2441,22 @@ where
...
 // private free-standing functions
 //
+
+/// Returns `true` if this entry is expired by its per-entry TTL.
+#[inline]
+fn is_expired_by_per_entry_ttl<K>(entry_info: &TrioArc<EntryInfo<K>>, now: Instant) -> bool {
+    if let Some(ts) = entry_info.expiration_time() {
+        if ts <= now {
+            return true;
+        }
```
tatsuya6502, line 2451:
```suggestion
        return ts <= now;
```

tatsuya6502 (commented):
Left some suggested changes.

tatsuya6502 (commented):
Merging.

PR #250, opened 2023-03-31
Fix the CI for MSRV 1.51

tatsuya6502:
When testing the MSRV, downgrade `tokio` and `futures-util` to v1.26.0 and v0.3.27 respectively, so that they will compile.


tatsuya6502 (commented):
Merging.

PR #256, opened 2023-04-23
Remove deprecated `unsync` cache

tatsuya6502:
- Remove the `unsync` cache that was marked as deprecated in v0.10.0.
- Update a dev dependency `env_logger` from v0.9 to v0.10.0.


tatsuya6502 (commented):
Merging.

PR #258, opened 2023-04-23
Add the `remove` method

### LINKED ISSUE #255: feature request: remove method
First, thanks for this awesome crate! It's been really helpful for us.

I was wondering if it would be possible to add a `remove` method to [moka](https://docs.rs/moka/0.10.2/moka/index.html)::[future](https://docs.rs/moka/0.10.2/moka/future/index.html)::[Cache](https://docs.rs/moka/0.10.2/moka/future/struct.Cache.html#) that removes a key from the cache, returning the value at the key if the key was previously in the cache.

Right now I am manually invalidating keys after calling `get()`, but a single method that could do both operations would be pretty handy.

tatsuya6502:
This PR adds the `remove` method to `future` and `sync` caches. It does the same to the `invalidate` method except returning a _clone_ of the value that has been discarded.

```rust
// moka::future::Cache

/// Discards any cached value for the key and returns a _clone_ of the value.
///
/// If you do not need to get the value that has been discarded, use the
/// [`invalidate`](#method.invalidate) method instead.
///
/// The key may be any borrowed form of the cache's key type, but `Hash` and `Eq`
/// on the borrowed form _must_ match those for the key type.
pub async fn remove<Q>(&self, key: &Q) -> Option<V>
where
    K: Borrow<Q>,
    Q: Hash + Eq + ?Sized;

// FYI, an existing `invalidate` method.
pub async fn invalidate<Q>(&self, key: &Q)
where
    K: Borrow<Q>,
    Q: Hash + Eq + ?Sized,
```

Fixes #255.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,16 +2,16 @@
 
 ## Version 0.11.0
 
-### Breaking Changes
-
-- (Will change the return type of the `invalidate` method from `()` to `Option<V>`)
-
 ### Added
 
-- Added support for per-entry expiration. ([#248][gh-pull-0248])
+- Added support for per-entry expiration ([#248][gh-pull-0248]):
     - In addition to the existing TTL and TTI (time-to-idle) expiration times that
       apply to all entries in the cache, the `sync` and `future` caches can now allow
       different expiration times for individual entries.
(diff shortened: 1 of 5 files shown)


tatsuya6502 (commented):
Merging.

PR #259, opened 2023-04-27
Fix the caches mutating a deque node through a `NonNull` pointer derived from a shared reference

tatsuya6502:
I found bugs in the `sync` and `future` caches; they were mutating an internal deque node through a `NonNull` pointer derived from a shared ref `&`. This PR fixes these bugs.

These bugs already existed from very early version of Moka but I could not find them until now. [Miri](https://github.com/rust-lang/miri) tests could have helped to find them, but we are not running Miri tests on the `sync_base` module who has the bugs. This is because the module uses some FFI calls that Miri does not support.

In commit https://github.com/moka-rs/moka/commit/03c75ff0ae7d39fc21f8d206ca0103ec59bbc7fb (included in this PR), I added a test to `deque` module to reproduce the bug and confirmed that Miri can catch it.

```console
$ cargo +nightly miri test deque

test common::deque::tests::peek_and_move_to_back ... error: Undefined Behavior: trying to retag from <506610> for Unique permission at alloc194760[0x0], but that tag only grants SharedReadOnly permission for this location
   --> /Users/tatsuya/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/non_null.rs:427:18
    |
427 |         unsafe { &mut *self.as_ptr() }
    |                  ^^^^^^^^^^^^^^^^^^^
    |                  |
    |                  trying to retag from <506610> for Unique permission at alloc194760[0x0], but that tag only grants SharedReadOnly permission for this location
    |                  this error occurs as part of retag at alloc194760[0x0..0x28]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <506610> was created by a SharedReadOnly retag at offsets [0x0..0x28]
   --> src/common/deque.rs:733:37
    |
733 |         unsafe { deque.move_to_back(NonNull::from(node1a)) };
    |                                     ^^^^^^^^^^^^^^^^^^^^^
    = note: BACKTRACE (of the first span):
    = note: inside `std::ptr::NonNull::<common::deque::DeqNode<std::string::String>>::as_mut::<'_>` at /Users/tatsuya/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/non_null.rs:427:18: 427:37
note: inside `common::deque::Deque::<std::string::String>::move_to_back`
   --> src/common/deque.rs:197:20
    |
197 |         let node = node.as_mut(); // this one is ours now, we can create an &mut.
    |                    ^^^^^^^^^^^^^
note: inside `common::deque::tests::peek_and_move_to_back`
   --> src/common/deque.rs:733:18
    |
733 |         unsafe { deque.move_to_back(NonNull::from(node1a)) };
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This PR fixes the bugs by not using a shared reference in these cases:

- Add `Deque::peek_front_ptr` method, which is similar to existing `peek_front` method. It returns a `NonNull` pointer to the front node, instead of a shared reference.
- Remove `DeqNode::next` method that returns a shared reference to the next node.
- Add `DeqNode::next_ptr` function that returns a `NonNull` pointers to the next node.
- Add `Deque::move_front_to_back` method that moves the front node to the back of the deque, instead of moving the node through a shared reference to the node.
- Modify the code in the `sync_base` module to use above methods when they need to mutate a `DeqNode`. 

This PR also adds the test mentioned above to `deque` module to ensure that the bugs are fixed.

I do not know if there were actual impacts of the bugs. I think the bugs were not triggered because the code never accesses the shared reference again after the mutable reference is created. Also the code is executed only when some corner cases are met.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@
     - Like the `invalidate` method, this method discards any cached value for the
       key, but returns a clone of the value.
 
+### Fixed
+
+- Fixed the caches mutating a deque node through a `NonNull` pointer derived from a
+  shared reference. ([#259][gh-pull-0259])
+
 ### Removed
 
 - Removed `unsync` cache that was marked as deprecated in [v0.10.0](#version-0100).
@@ -642,6 +647,7 @@ The minimum supported Rust version (MSRV) is now 1.51.0 (2021-03-25).
 [gh-issue-0034]: https://github.com/moka-rs/moka/issues/34/
 [gh-issue-0031]: https://github.com/moka-rs/moka/issues/31/
 
+[gh-pull-0259]: https://github.com/moka-rs/moka/pull/259/
 [gh-pull-0251]: https://github.com/moka-rs/moka/pull/251/
 [gh-pull-0248]: https://github.com/moka-rs/moka/pull/248/
 [gh-pull-0216]: https://github.com/moka-rs/moka/pull/216/
--- a/src/common/concurrent/deques.rs
+++ b/src/common/concurrent/deques.rs
@@ -194,3 +194,5 @@ impl<K> Deques<K> {
         }
     }
 }
+
+// TODO: Add tests and run Miri with them.
--- a/src/common/deque.rs
+++ b/src/common/deque.rs
@@ -335,6 +335,8 @@ impl<T> Deque<T> {
 
 #[cfg(test)]
 mod tests {
+    use std::ptr::NonNull;
+
     use super::{CacheRegion::MainProbation, DeqNode, Deque};
 
     #[test]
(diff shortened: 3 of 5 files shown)


tatsuya6502 (commented):
Merging.

tatsuya6502:
> I do not know if there were actual impacts of the bugs. I think the bugs were not triggered because the code never accesses the shared reference again after the mutable reference is created. Also the code is executed only when some corner cases are met.

I was wrong. A Moka v0.9.6 user reported segmentation fault and a core dump implied that it was caused by this bug:
https://github.com/moka-rs/moka/issues/281#issuecomment-1616445598

I backported this fix to v0.9.8 and also plan to backport to v0.10.x.

PR #260, opened 2023-04-29
Per-entry expiration. (Add more doc and tests)

tatsuya6502:
This PR adds more rustdoc and unit tests for per-entry expiration #248.


tatsuya6502 (commented):
Merging.

Issue #269, opened 2023-05-27
CI: Miri test causes a complie error with Rust nightly-2023-05-27


tatsuya6502:
https://github.com/moka-rs/moka/actions/runs/5098288414/jobs/9165346113

```console
/home/runner/.cargo/bin/rustup toolchain install nightly --component miri --allow-downgrade
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
info: latest update on 2023-05-27, rust version 1.71.0-nightly (1a5f8bce7 2023-05-26)
...

  nightly-x86_64-unknown-linux-gnu installed - rustc 1.71.0-nightly (1a5f8bce7 2023-05-26)
...

Run actions-rs/cargo@v1
/home/runner/.cargo/bin/cargo miri test deque
Running `"rustup" "component" "add" "rust-src"` to install the `rust-src` component for the selected toolchain.
info: downloading component 'rust-src'
info: installing component 'rust-src'
Preparing a sysroot for Miri (target: x86_64-unknown-linux-gnu)... done
    Updating crates.io index
 Downloading crates ...
...
  Compiling proc-macro2 v1.0.59
...
error: to use a constant of type `core::any::TypeId` in a pattern, `core::any::TypeId` must be annotated with `#[derive(PartialEq, Eq)]`
Error:   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/value-bag-1.0.0-alpha.9/src/internal/cast/primitive.rs:43:33
   |
43 |                                   $const_ident => |v| Some(Internal::from(unsafe { &*(*v as *const Self as *const $ty) })),
   |                                   ^^^^^^^^^^^^
...
71 | /         to_internal![
72 | |             usize: (USIZE, OPTION_USIZE),
73 | |             u8: (U8, OPTION_U8),
74 | |             u16: (U16, OPTION_U16),
...  |
96 | |             String: (STRING, OPTION_STRING),
97 | |         ];
   | |_________- in this macro invocation
   |
   = note: the traits must be derived, manual `impl`s are not sufficient
   = note: see https://doc.rust-lang.org/stable/std/marker/trait.StructuralEq.html for details
   = note: this error originates in the macro `to_internal` (in Nightly builds, run with -Z macro-backtrace for more info)
```

```console
$ cargo tree -i value-bag
value-bag v1.0.0-alpha.9
└── log v0.4.17
 ...
```

tatsuya6502:
[...]

PR #270, opened 2023-05-27
CI: Workaround a CI failure for Miri test by using a specific version of Rust nightly

tatsuya6502:
To workaround the issue described by #269, use a specific version of Rust nightly for Miri test for now. We are going to use `nightly-2023-05-24`.

We will revert this change when the original issue is addressed.


tatsuya6502 (commented):
Merging.

PR #271, opened 2023-05-27
Tweak some examples

tatsuya6502:
- Bump the Moka version to v0.11.1.
- Reorder some cache operations in the `eviction_listener` example and add some comments.
- Apply `cargo fmt`.
- Update the change log.


tatsuya6502 (commented):
Merging.

Issue #257, opened 2023-04-23
Use `std::thread::available_parallelism()` instead of `num_cpus` dependency


tatsuya6502:
We might want to replace a dependency `num_cpus` crate with `std::thread::available_parallelism()` once we update `moka`'s minimum supported Rust version (MSRV) to Rust 1.64.

- `std::thread::available_parallelism()` function was added to Rust 1.59.0.
    - But Linux cgroup v2 support was added to the function in Rust 1.61.0.
    - And Linux cgroup v1 support was added to the function in Rust 1.64.0.

See https://github.com/rayon-rs/rayon/pull/937 for details.

Without Linux cgroup v1/v2 support, `available_parallelism()` will ignore the CPU quota provided by a Linux container runtime such as Docker.

[Rust 1.64.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1640-2022-09-22) was released on September 22, 2022. It was 7 months ago from now and our MSRV policy already allows Rust 1.64.0.

https://github.com/moka-rs/moka/#minimum-supported-rust-versions

> It will keep a rolling MSRV policy of at least 6 months.

As of `moka` v0.10, it uses `num_cpus` to decide the number of global worker threads. In the future, we will remove the worker threads, but we will start to use it to decide the number of the internal hash table segments.

PR #275, opened 2023-06-03
Bump the minimum supported Rust version (MSRV) to 1.65 (Nov 3, 2022)

tatsuya6502:
Change the MSRV from 1.60 to 1.65. Rust 1.65 was released on Nov 3, 2022, seven months ago.

We need Rust 1.65 for the followings:

- #257
    - `std::thread::available_parallelism()` requires 1.64 in order to work reliably on some Linux platforms.
    - We can remove `num_cpus` crate.
- #265
    - Not mandatory, but it uses `if let — else`, which was introduced in 1.65.

Also this PR bumps the `moka` version to v0.11.2.


tatsuya6502 (commented):
Merging.

PR #277, opened 2023-06-06
Remove `num_cpus` crate from the dependency

### LINKED ISSUE #257: Use `std::thread::available_parallelism()` instead of `num_cpus` dependency

tatsuya6502:
This PR replaces `num_cpus::get()` with `std::thread::available_concurrency()` introduced in Rust 1.59.0. (Fixes #257)

NOTE: `available_concurrency()` requires Rust 1.64.0 or newer to work reliably in some Linux container environment. See the descriptions of #257.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,11 @@
 ## Version 0.11.2
 
 Bumped the minimum supported Rust version (MSRV) to 1.65 (Nov 3, 2022).
-([#275][gh-issue-0275])
+([#275][gh-pull-0275])
 
+### Removed
+
+- Removed `num_cpus` crate from the dependency. ([#277][gh-pull-0277])
 
 ## Version 0.11.1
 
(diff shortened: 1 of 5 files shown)


tatsuya6502 (commented):
Merging.

PR #291, opened 2023-07-28
Fix Clippy warnings for v0.11.x

tatsuya6502:
- Fix Clippy warnings.
    - clippy 0.1.72 (a47f796a365 2023-07-22)
- Add a cfg flag called `clippy_beta` to allow a lint when Clippy beta is used.
    - This is needed because the same lint is not available in Clippy stable.


tatsuya6502 (commented):
Merging.

PR #292, opened 2023-07-29
Fix Cilppy warnings and MSRV CI for v0.10.x

tatsuya6502:
## Fix Clippy warnings

- clippy 0.1.72 (a47f796a365 2023-07-22)
- Add a cfg flag called `clippy_beta` to allow a lint when Clippy beta is used.
    - This is needed because the same lint is not available in Clippy stable.

## Fix CI for the MSRV

- When testing the MSRV (1.60.0) of `moka` 0.10.x, downgrade `tempfile` to v3.6.0 so it will compile.

### DIFF
--- a/.ci_extras/pin-crate-vers-msrv.sh
+++ b/.ci_extras/pin-crate-vers-msrv.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+set -eux
+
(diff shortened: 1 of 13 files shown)


tatsuya6502 (commented):
Merging.

PR #295, opened 2023-08-03
Fix a bug in sync caches where memory usage kept increasing when the eviction listener is set (v0.11.x)

tatsuya6502:
This PR fixes a bug in `sync::Cache` and `sync::SegmentedCache`, which causes eviction listener's key-level locks not to be removed after sending notifications. This happened only when the `Immediate` notification delivery mode is used for the eviction listener. (It is the default mode)  Another mode, `Queued` mode, is unaffected as it does not use the key locks.

These key locks are stored in an internal `moka::cht::SegmentedHashMap` with `Arc<K>` as their keys. So this bug caused the `SegmentedHashMap` to hold all keys `K` that has been inserted to the cache. It will continuously increasing memory usage until the cache is dropped.

Tested the fix by adding a check to unit test if the `SegmentedHashMap` is empty after sending all notifications.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   increasing when the eviction listener was set with the `Immediate` delivery mode.
   ([#295][gh-pull-0295])
 
+
 ## Version 0.11.2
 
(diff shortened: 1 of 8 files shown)


tatsuya6502 (commented):
Merging.

PR #294, opened 2023-08-01
Remove the thread pool from `future::Cache`

tatsuya6502:
This PR changes the followings on `future::Cache`:

- Remove the thread pool from `future::Cache`. The background threads will be no longer spawned.
- Change the `notification::DeliveryMode` for eviction listener from `Queued` to `Immediate`.
    - #228

For more details, see the [migration guide at commit #16b4f899](https://github.com/moka-rs/moka/blob/16b4f899a27a1e8d89504c339e255aa32cab7dfa/MIGRATION-GUIDE.md).

PR #309, opened 2023-08-23
Improve async cancellation safety of `future::Cache` (v0.12.0-beta.X)

tatsuya6502:
This PR improves async cancellation safety of cache write methods `insert`, `get_with`, `invalidate` and `remove`. It is for `future::Cache` in v0.12.0-beta.1.

## The Problem

https://github.com/moka-rs/moka/pull/294#issuecomment-1687337263

> I was reviewing the codes and found that cache write operations (such as `insert` and `get_with`) may not record a write operation log if the caller (enclosing future) has been cancelled. This will leave a newly inserted cached entry not managed by the cache policies (LFU filter, LRU queue, etc.). When this happens, the entry will never be evicted by the policy, while it can be read by `get`, replaced by other `insert`, or removed by `invalidate`.
> 
> This issue already existed in very early version of moka. But v0.12.0 will increase the chance of this issue to happen as it now supports for calling user supplied async eviction listener.
> 
> A common solution for this would be write-ahead-log, which to record the write op log ahead of the actual insert/update operation. However, we will not be able to do this because it is impossible to know the exact operation we are going to perform (_insert_ or _update_ to the internal concurrent hash table (cht)). Our lock-free cht allows multiple threads to try to insert and remove the same key at the same time, without taking locks. We will know our insert attempt turned out to be an insert or update only _after_ we succeeded.

You will be affected by the above if you use `async_eviction_listener` _and_ do [async cancellation](https://blog.yoshuawuyts.com/async-cancellation-1/). For example, some sever frameworks such as Axum do async cancellation for you; request handling will be cancelled when the client connection is dropped.

## The Solution

This PR addresses the above problem by saving the interrupted tasks to a mpmc channel when async cancellation happens on the following methods:

- Some of the cache write methods: `insert`, `get_with`, `invalidate` and `remove`

These interrupted tasks will be resumed/retried when one of the following methods is called:

- The cache write methods listed above.
- `get` method
- `run_pending_tasks` method

The following interrupted tasks will be saved and resumed/retried:

- If an eviction listener is configured, and an async cancellation happens while one of those cache write methods listed above is notifying the eviction listener for an entry replacement/removal, the incomplete notification `Future` should be saved.
- An async cancellation happens while one of those cache write methods listed above tries to send a write op log (`WriteOp` enum) to an internal channel but the channel is full, the `WriteOp` should be saved.

## A Limitation

`future::Cache` v0.12.0 or later does `Immediate` notification delivery mode, which guarantees to preserve the order of events on a key. However, if the eviction listener is interrupted by async cancellation and then resumed, the order may not be preserved. 

## Testing the Solution

Added the following unit tests to verify those interrupted tasks are resumed/retried:

- `cancel_future_while_calling_eviction_listener`
- `cancel_future_while_scheduling_write_op`

### DIFF
--- a/MIGRATION-GUIDE.md
+++ b/MIGRATION-GUIDE.md
@@ -242,7 +242,6 @@ free some memory), you do not need to call `run_pending_tasks` method.
    - To enable it, see [Enabling the thread pool](#enabling-the-thread-pool) for more
      details.
 
-
 #### Enabling the thread pool
 
 To enable the thread pool, do the followings:
--- a/MIGRATION-GUIDE.md
+++ b/MIGRATION-GUIDE.md
@@ -146,7 +146,7 @@ let cache = Cache::builder()
 
 `async_eviction_listener` takes a closure that returns a `Future`. If you need to
 `.await` something in the eviction listener, use this method. The actual return type
-of the closure is `future::ListenerFuture`, which is a type alias of
+of the closure is `notification::ListenerFuture`, which is a type alias of
 `Pin<Box<dyn Future<Output = ()> + Send>>`. You can use the `boxed` method of
 `future::FutureExt` trait to convert a regular `Future` into this type.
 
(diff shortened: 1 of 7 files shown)


tatsuya6502 (commented):
Merging.

PR #310, opened 2023-08-25
rustfmt and clippy (v0.12.x)

tatsuya6502:
- Apply rustfmt for `if let ... else`.
- Remove allow `clippy::arc_with_non_send_sync` as the false positives are fixed in Clippy.


tatsuya6502 (commented):
Merging.

PR #316, opened 2023-09-03
Remove the thread pool from `sync` caches

tatsuya6502:
This PR changes the followings on `sync::Cache` and `sync::SegmentedCache`:

- Remove a crate feature `sync` from the default features. Now `sync` caches are disabled by default.
- Remove the thread pool. The background threads will be no longer spawned.
- Remove `notification::DeliveryMode` enum for eviction listener.
    - Now caches work as if `DeliveryMode::Immediate` is specified.

For more details, see the [migration guide](https://github.com/moka-rs/moka/blob/main/MIGRATION-GUIDE.md).


tatsuya6502 (commented):
Merging.

PR #323, opened 2023-09-09
Better async waiting for when the write op channel is full

tatsuya6502:
This PR replaces the async `may_yield` function introduced at `v0.12.0-beta.1` with a more efficient implementation. It uses an `async_lock::RwLock` between a user async task, that is currently running internal `do_run_pending_tasks`, and other user async tasks.

This will be more CPU efficient than `may_yield` because it makes some of the user async tasks to wait for reader locks, rather than keeping spinning, so that the async runtime scheduler can run other async tasks if there are any.

This PR also bumps the version to `v0.12.0-beta.3`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "moka"
-version = "0.12.0-beta.2"
+version = "0.12.0-beta.3"
 edition = "2018"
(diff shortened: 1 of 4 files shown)


tatsuya6502 (commented):
Merging.

PR #324, opened 2023-09-09
Get the last modified time for upsert after acquiring key level lock

### LINKED ISSUE #322: An internal `do_insert_with_hash` method gets the current `Instant` too early when eviction listener is enabled
Relates to:

- #321 

As of moka v0.12.0-beta.2, an internal method `do_insert_with_hash` gets the `ts` (the base `Instant`, or the time that upsert is about to be performed at) before acquiring the key-level lock for notifying to the eviction listener. This could make the base instant much earlier than the actual time that the entry was upserted at, because it can take some time until the lock becomes available. 

Therefore, we should change it to get the base instant _after_ acquiring the lock.

Later, the base instant is passed to an `Expiry` method as the `current_time`, and when the expiry method returns, it is used to calculate the expiration time from the returned expiry duration. (`expires_at = current_time + returned_duration`). The base instant is also used by the time-to-live and time-to-idle that can be set at the cache creation time.

https://github.com/moka-rs/moka/blob/a03488b02d4d3d0567e6974ed424d7cc461cff8b/src/sync_base/base_cache.rs#L483-L492

https://github.com/moka-rs/moka/blob/a03488b02d4d3d0567e6974ed424d7cc461cff8b/src/future/base_cache.rs#L477-L490

## Action Items

- [x] Fix v0.12.x via #324.
- [x] Release v0.12.0.
- [ ] Fix v0.11.x.
- [ ] Release v0.11.4.

tatsuya6502:
- Fixes #322
- Update the `Expiry` documentation for clarity.

This is for the `main` branch (v0.12.0). I will back-port this to the branch for v0.11.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ Please read the [MIGRATION-GUIDE.md][migration-guide-v012] for more details.
   caches ([#316][gh-pull-0316]).
 - Improved async cancellation safety of `future::Cache`. ([#309][gh-pull-0309])
 
+### Fixed
+
+- Fixed a bug that an internal `do_insert_with_hash` method gets the current
+  `Instant` too early when eviction listener is enabled. ([#322][gh-issue-0322])
+
 
 ## Version 0.11.3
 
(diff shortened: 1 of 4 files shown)


tatsuya6502 (commented):
Merging.

PR #331, opened 2023-10-02
Micro-optimize ValueInitializer

Swatinem:
This has the following changes/optimizations:
- Avoids a needless incref/decref on `waiters`.
- Avoids unconditionally cloning the waiter key by moving it to the `WaiterGuard`.
- Use an `Option` in the `WaiterGuard` instead of a manual bool.
- Move the own `waiter` allocation out of the loop.
- Un-indent the loop by using an let-else.
- This also clarifies that no looping happens when our waiter was inserted.

Based on top of #330, I will rebase once that lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Moka Cache &mdash; Change Log
 
+## Version 0.12.1
+
+### Fixed
+
+- Fixed memory leak in `future::Cache` when `get_with()`, `entry().or_insert_with()`,
+  and friend methods are used ([#329][gh-issue-0329]).
+    - This bug was introduced in `v0.12.0`. Versions earlier than `v0.12.0` do not
+      have this bug.
+
+
 ## Version 0.12.0
 
 > **Note**
@@ -702,6 +712,7 @@ The minimum supported Rust version (MSRV) is now 1.51.0 (Mar 25, 2021).
 [gh-Swatinem]: https://github.com/Swatinem
 [gh-tinou98]: https://github.com/tinou98
 
+[gh-issue-0329]: https://github.com/moka-rs/moka/issues/329/
 [gh-issue-0322]: https://github.com/moka-rs/moka/issues/322/
 [gh-issue-0255]: https://github.com/moka-rs/moka/issues/255/
 [gh-issue-0252]: https://github.com/moka-rs/moka/issues/252/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "moka"
-version = "0.12.0"
+version = "0.12.1"
 edition = "2018"
 # Rust 1.65 was released on Nov 3, 2022.
 rust-version = "1.65"
--- a/src/future/base_cache.rs
+++ b/src/future/base_cache.rs
@@ -498,7 +498,7 @@ where
 
         // TODO: Instead using Arc<AtomicU8> to check if the actual operation was
         // insert or update, check the return value of insert_with_or_modify. If it
-        // is_some, the value was inserted, otherwise the value was updated.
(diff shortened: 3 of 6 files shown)


Swatinem:
Some other interesting observations while refactoring this code:
- The `ignore_if` closure is never used/executed if we get a fresh value from an existing waiter.
- Now that I understand the `TypeId` usage, it is clear to me now that it is used to facilitate the `try_` and `optionally_` family of accessors.
  However that also means that mixing different kinds of accessors means that they are concurrently executing their respective `init` closure. That also means that you might get inconsistent results.

tatsuya6502:
Very nice! Thank you.

> The `ignore_if` closure is never used/executed if we get a fresh value from an existing waiter.

I tried to remove `Arc<Mutex<..>>` from `ignore_if` before, but gave up as the compiler complained the resulting `Future` of `try_init_or_read` was not `Send` and `Sync`. I do not remember details but maybe I had `&mut ignore_if` instead of `mut ignore_if` because it was in the loop? I am glad you found the solution!

I will run some tests after work to check if your change works as expected.

tatsuya6502:
> Based on top of https://github.com/moka-rs/moka/pull/330, I will rebase once that lands.

Thanks. I merged #330 but you do not have to rebase.

I changed the target branch of this pull request (PR) to `fix-get-with-in-future-cache` and it eliminated the extra commit histories and diffs. Then I changed the target back to `main`, but my colleague said I did not have to do it. He said: when #330 was merged into `main`, the target branch of this PR should have automatically been changed to `main`.

So next time, you will simply create a PR targeting another topic branch rather than `main`. Then GitHub will take care of the rest; it will automatically change the target branch of the PR to `main` when the target branch is merged in to `main`.

tatsuya6502:
> * Now that I understand the `TypeId` usage, it is clear to me now that it is used to facilitate the `try_` and `optionally_` family of accessors.
>   However that also means that mixing different kinds of accessors means that they are concurrently executing their respective `init` closure. That also means that you might get inconsistent results.

You are right. Mixing different `init` closures will lead unexpected result.

I will open an issue for it. I guess we can solve it by adding the `TypeId` of `init` closure/future to the waiter key? The compiler will create an opaque type for `init` closure/future, and if you have two different `init` codes, I think they will have different `TypeId`s.

Swatinem:
> I tried to remove `Arc<Mutex<..>>` from `ignore_if` before, but gave up as the compiler complained the resulting `Future` of `try_init_or_read` was not `Send` and `Sync`. I do not remember details but maybe I had `&mut ignore_if` instead of `mut ignore_if` because it was in the loop? I am glad you found the solution!

I added an explicit test to make sure all the futures in the public API are `Send` at least.
However they are not `Sync`, neither on `master`.

While going through all the public API, I noticed that maybe this function is not supposed to be public?
https://docs.rs/moka/latest/moka/future/struct.Cache.html#method.invalidate_with_hash

tatsuya6502:
> I added an explicit test to make sure all the futures in the public API are `Send` at least. However they are not `Sync`, neither on `master`.

Thanks!

> While going through all the public API, I noticed that maybe this function is not supposed to be public? https://docs.rs/moka/latest/moka/future/struct.Cache.html#method.invalidate_with_hash

Ah. You are right. Can you do me a favor? Can you please change it to private in this PR?

Swatinem:
> Can you please change it to private in this PR?

Done. While this is *in theory* a breaking change, I don’t think anyone will really care :-)

tatsuya6502 (approved):
Thanks. It looks good to me.

I am running mokabench with different command line arguments (on the commit https://github.com/moka-rs/moka/pull/331/commits/edcc240a969863d91c6ca285b1bc83facf4b1281). It has been 1.5 hours since started and I have not seen any issues.

Issue #333, opened 2023-10-11
With Rust 1.73.0, some unit tests started to fail for `mips-unknown-linux-musl` target


tatsuya6502:
Rust 1.73.0 has been released and some unit tests started to fail for 32-bit `mips-unknown-linux-musl` target.

Note that the CI is using QEMU's user space emulator, so emulation may not be accurate. However, as long as I can remember, these tests never failed with older versions of Rust.

https://github.com/moka-rs/moka/actions/runs/6436542687/job/17587394644

```console
failures:

---- common::timer_wheel::tests::test_advance stdout ----
thread 'common::timer_wheel::tests::test_advance' panicked at src/common/timer_wheel.rs:667:22:
Expected an expired entry. Got Rescheduled(EntryInfo { key_hash: KeyHash { key: 1, hash: 1 }, is_admitted: false, is_dirty: true, last_accessed: AtomicInstant { instant: RwLock { data: Some(Instant(Instant { tv_sec: 434, tv_nsec: 814957001 })) } }, last_modified: AtomicInstant { instant: RwLock { data: Some(Instant(Instant { tv_sec: 434, tv_nsec: 814957001 })) } }, expiration_time: AtomicInstant { instant: RwLock { data: Some(Instant(Instant { tv_sec: 439, tv_nsec: 814957001 })) } }, policy_weight: 0 })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- common::timer_wheel::tests::test_bucket_indices stdout ----
thread 'common::timer_wheel::tests::test_bucket_indices' panicked at src/common/timer_wheel.rs:586:9:
assertion `left == right` failed
  left: (0, 3)
 right: (0, 63)

---- sync::cache::tests::time_to_idle_by_expiry_type stdout ----
thread 'sync::cache::tests::time_to_idle_by_expiry_type' panicked at src/sync/cache.rs:2625:9:
assertion `left == right` failed
  left: 2
 right: 1


failures:
    common::timer_wheel::tests::test_advance
    common::timer_wheel::tests::test_bucket_indices
    sync::cache::tests::time_to_idle_by_expiry_type

test result: FAILED. 83 passed; 3 failed; 3 ignored; 0 measured; 0 filtered out; finished in 30.27s

```

tatsuya6502:
[...]

PR #334, opened 2023-10-14
CI: Update Linux cross compile tests (v0.12.x)

tatsuya6502:
Update Linux cross compile tests (linux-cross) on the `main` branch (for v0.12.x).

- Disable fail-fast strategy for linux-cross.
- Stop using actions-rs GH actions in linux-cross:
    - Stop using `actions-rs/toolchain` and `actions-rs/cargo`.
    - Start to use `dtolnay/rust-toolchain` and `taiki-e/install-action`.
    - Upgrade `actions/checkout` from v2 to v4.

Note that the following tests are failing due to a Rust compiler/LLVM bug for MIPS targets. (See #333)

- linux-cross (mips-unknown-linux-musl)
- linux-cross (mipsel-unknown-linux-musl)

### DIFF
--- a/.github/workflows/LinuxCrossCompileTest.yml
+++ b/.github/workflows/LinuxCrossCompileTest.yml
@@ -21,6 +21,8 @@ jobs:
   linux-cross:
     runs-on: ubuntu-latest
     strategy:
+      # Continue running other jobs in the matrix even if one fails.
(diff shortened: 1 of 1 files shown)


tatsuya6502 (commented):
Merging.

PR #339, opened 2023-11-09
Update to 2021 edition, bless tests

nyurik:
* Move to edition 2021, and remove redundant `use` in test
* run tests with `TRYBUILD=overwrite`

PR #340, opened 2023-11-09
Inline format args

nyurik:
Makes the code a bit easier to read. Plus a few unneeded commas.

Some of these require migration to 2021 edition first (another PR)


tatsuya6502:
Thank you very much for submitting a PR!

I believe we can ignore the failing tests. (See https://github.com/moka-rs/moka/pull/339#issuecomment-1804790833)

tatsuya6502:
Memo: Inline format args were introduced in [Rust 1.58.0](https://releases.rs/docs/1.58.0/). Moka v0.12.1's MSRV is 1.65.0.

tatsuya6502 (approved):
Wow, there were lots of places to fix. Thank you for doing this!

PR #343, opened 2023-11-10
A slightly better panics

nyurik:
Does this pattern make more sense?  Wasn't certain, so made a separate PR


tatsuya6502 (approved):
They look good. Thanks!

PR #348, opened 2023-11-19
Prevent timing issues causing inconsistencies between concurrent hash table and policy data structures

### LINKED ISSUE #345: Entry count never reaches zero even when calling run_pending_tasks() in a loop [moka future 0.12]
Trying to convert the eviction listener example from a `sync` version to a `future` version, I see some weird behavior. Depending on the presence of a _second insertion_ of the same key, the `entry_count()` will never reach zero, even after calling `run_pending_tasks()` in a loop. See the example below. Uncommenting the line number 48 (the line after the line with the triple question mark) will hold the `entry_count()` above zero forever, causing the loop to continue forever.

Some remarks:
1. Is this expected behavior?
2. Would it be an idea to let `run_pending_tasks()` return a `true` value if it is finished or return a `false` if more than 500 items were processed and it is still not finished. (Finished from the perspective of e.g. the timestamp of calling the function, later insertions can be ignored for this judgement).
3. I wrote this example to study the possibility to use the information `Arc::strong_count()` in the eviction listener, see [discussion 344](https://github.com/moka-rs/moka/discussions/344) where I want to avoid orphaned entries in the `Btree` with shared `Arc`'s.

```rust
use moka::future::Cache;
use std::collections::BTreeMap;
use std::sync::Arc;
use std::thread::sleep;
use std::time::Duration;

#[tokio::main]
async fn main() {
    // Store user names in a B-tree.
    let mut tree = BTreeMap::new();
    let arc1 = Arc::new(String::from("Alice"));
    let arc2 = Arc::new(String::from("Bob"));
    let arc3 = Arc::new(String::from("Charlie"));
    let arc4 = Arc::new(String::from("Dave"));
    let arc5 = Arc::new(String::from("Eve"));
    let arc6 = Arc::new(String::from("Faythe"));
    tree.insert(1, arc1.clone());
    tree.insert(2, arc2.clone());
    tree.insert(3, arc3.clone());
    tree.insert(4, arc4.clone());
    tree.insert(5, arc5.clone());
    tree.insert(6, arc6.clone());

    // Make an artificially small cache and 1-second ttl to observe eviction listener.
    let ttl = 1;
    {
        let cache = Cache::builder()
            .max_capacity(2)
            .time_to_live(Duration::from_secs(ttl))
            .eviction_listener(|key, value, cause| {
                println!("Evicted ({:08X},{:?}) because {:?}", *key, value, cause)
            })
            .build();
        // Create some simple CRC-32 session keys with Bash:
        //   for ((i = 1; i < 7 ; i++)); do rhash <(echo "$i")|tail -1; done
        cache.insert(0x6751FC53, arc1.clone()).await;
        cache.insert(0x4C7CAF90, arc2.clone()).await;
        cache.insert(0x55679ED1, arc3.clone()).await;
        cache.insert(0x1A260816, arc4.clone()).await;
        cache.insert(0x033D3957, arc5.clone()).await;
        cache.insert(0x28106A94, arc6.clone()).await;
        // Due to race condition spilled over maybe evicted twice by cause
        // Replaced and Size.
        cache.insert(0x4C7CAF90, arc2.clone()).await;
        // With 1-second ttl, keys 0 and 1 will be evicted if we wait long enough.
        sleep(Duration::from_secs(ttl));
        println!("Wake up!");
        // UNCOMMENTING THE NEXT LINE WILL BLOCK THE run_pending_tasks() LOOP LATER ON?        <------ ???
        //cache.insert(0x55679ED1,arc3.clone()).await;

        // Remove from cache and return value:
        if let Some(v) = cache.remove(&0x55679ED1).await {
            println!("Removed: {v}")
        };
        // Or remove from cache without returning the value.
        cache.invalidate(&0x1A260816).await;

        // invalidate_all() removes entries using a background thread, so there will
        // be some delay before entries are removed and the eviction listener is
        // called. If you want to remove all entries immediately, call
        // run_pending_tasks() method repeatedly like the loop below.
        cache.invalidate_all();
        loop {
            // Synchronization is limited to at most 500 entries for each call.
            // Check if all is done. Calling entry_count() requires calling
            // run_pending_tasks() first!
            if cache.entry_count() == 0 {
                break;
            }
        }

        cache.insert(0x28106A94, arc6.clone()).await;
        // When cache is dropped eviction listener is not called. Either call
        // invalidate_all() or wait longer than ttl.
        sleep(Duration::from_secs(ttl + 1));
    } // cache is dropped here.

    println!("Cache structure removed.");
    sleep(Duration::from_secs(1));
    println!("Exit program.");
}

```

tatsuya6502:
## Descriptions

Fixes #345.

This pull request (PR) prevents timing issues in writes that cause inconsistencies between the cache's internal data structures such as the concurrent hash table and access-order queue.

## Example

The following new test case reproduces this issue in `v0.12.0` and `v0.12.1`:

- [`test_race_between_updating_entry_and_processing_its_write_op`](https://github.com/moka-rs/moka/pull/348/files#diff-338f772312880199137f8bbb508f15dd7f83457565cc983b85043cddd71b2f24)

1. Create a `Cache` with `max_capacity = 2` and `time_to_idle = 1 second`.
2. Insert key `a`.
3. Insert key `b`.
4. Insert key `c`.
    - → When pending tasks are processed (at step 6), `c` will be evicted as the cache is full.
5. Pause for 2 seconds.
    - → When pending tasks are processed (at step 6), `a` and `b` should expire.
6. Insert key `c` again.
    - → The pending tasks up to step 4 should be processed implicitly because more than 0.3 seconds have passed after creating the `Cache`.
7. Remove key `c`.
8. Pause for 2 seconds.
9. Call `run_pending_tasks`.
    - → The pending tasks from step 6 and 7 should be processed.
10. Get the `entry_count`.

### Expected

- At step 7, the value inserted at step 6 should be returned.
- At step 10, `0` should be returned.

## Actual

- At step 7, no value was returned.
- At step 10, `1` was be returned.

## Cause

When processing pending tasks, `Cache` did not check if more than one pending tasks exist for the same key.

In the above example, key `c` had two pending tasks at step 6; first one from step 4 (when cache is full) and second one from step 6 (when cache has enough room).

At step 6, pending tasks up to step 4 should be processed:

1. Write operation log for inserting key `a` is processed:
    - Key `a` is admitted and added to the access-order queue (the LRU deque).
2. Write operation log for inserting key `b` is processed:
    - Key `b` is admitted and added to the LRU.
3. Write operation log for inserting key `c` (at step 4) is processed:
    - It is evicted because the `Cache` is full, and removed from the concurrent hash table (CHT).
    - This removes the key-value of `c` (second insertion at step 6) from the CHT (**wrong**).
5. Key `a` is considered expired:
    - It is removed from both the CHT and LRU.
6. Key `b` is considered expired:
    - It is removed from both the CHT and LRU.

Then, at step 9, pending tasks from step 6 and 7 should be processed:

1. Write operation log for inserting key `c`  (at step 6) is processed:
    - Key `"c"` is admitted and added to the LRU.
    - (But there is no key-value for `c` in the CHT)
2. There is no write operation log for removing key `c` (at step 7).
    - This is because key-value for `c` did not exist at step 7.
3. Key `c` is considered expired:
    - Key `c` is supposed to be removed from both the CHT and LRU, but it will not be, because it does not exist in the CHT.

This issue was already present in `v0.11.x` and older versions, but were less likely to occur because the window for the timings was much smaller than `v0.12.0` and `v0.12.1` because these older versions used background threads to periodically process pending tasks.

## Fixes

When processing write operation logs in pending tasks, check if more than one logs exist for the same key:

- To achieve it, add two `u16` counters to each key in the CHT to track the following generations (versions):
    - `entry_gen`: Will be incremented on every insertion, update and removal of the key.
        - The write operation log carries the `entry_gen` after the increment.
    - `policy_gen`: Will be updated on every time the write operation log is processed for the key.
- When processing a write operation log, check if the `entry_gen` in the log equals to the current `entry_gen` of the key.
    - If yes, and the key is not admitted, remove the key from the CHT.
    - If no, and the key is not admitted, do nothing for the key. (This solves the problem in the above example)

After processing write operation logs, `Cache` will process other pending tasks such as evicting expired entries. This PR also incudes various small improvements in these steps by checking if each key still has pending write operation logs. This is done by calling `is_dirty` method, which returns `true` when `entry_gen != policy_gen`.

### Checklist

Add check if `entry_gen` in the write operation log equals to the current `entry_gen` of the key:

- `future::base_cache::Internal` struct
    - [x] `handle_upsert` method
- `sync::base_cache::Internal` struct 
    - [x] `handle_upsert` method

Add `is_dirty` checks to the following internal methods:

- `future::base_cache::Internal` struct
    - [x] `evict_expired_entries_using_timers` method
    - [x] `remove_expired_ao` method
    - [x] `remove_expired_wo` method
    - [x] `invalidate_entries` method
    - [x] `evict_lru_entries` method
- `sync::base_cache::Internal` struct 
    - [x] `handle_upsert` method
    - [x] `evict_expired_entries_using_timers` method
    - [x] `remove_expired_ao` method
    - [x] `remove_expired_wo` method
    - [x] `invalidate_entries` method
    - [x] `evict_lru_entries` method

### DIFF
--- a/src/common/concurrent/entry_info.rs
+++ b/src/common/concurrent/entry_info.rs
@@ -191,12 +191,12 @@ mod test {
         };
 
         let expected_sizes = match (arch, is_quanta_enabled) {
-            (Linux64, true) => vec![("1.51", 48)],
-            (Linux32, true) => vec![("1.51", 48)],
-            (MacOS64, true) => vec![("1.62", 48)],
-            (Linux64, false) => vec![("1.66", 96), ("1.60", 120)],
-            (Linux32, false) => vec![("1.66", 96), ("1.62", 120), ("1.60", 72)],
-            (MacOS64, false) => vec![("1.62", 96)],
+            (Linux64, true) => vec![("1.51", 56)],
+            (Linux32, true) => vec![("1.51", 56)],
+            (MacOS64, true) => vec![("1.62", 56)],
+            (Linux64, false) => vec![("1.66", 104), ("1.60", 128)],
+            (Linux32, false) => vec![("1.66", 104), ("1.62", 128), ("1.60", 80)],
+            (MacOS64, false) => vec![("1.62", 104)],
         };
 
         let mut expected = None;
--- a/src/common/concurrent/entry_info.rs
+++ b/src/common/concurrent/entry_info.rs
@@ -1,4 +1,4 @@
-use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU16, AtomicU32, Ordering};
 
 use super::{AccessTime, KeyHash};
 use crate::common::{concurrent::atomic_time::AtomicInstant, time::Instant};
@@ -10,10 +10,12 @@ pub(crate) struct EntryInfo<K> {
     /// cache. When `false`, it means the entry is _temporary_ admitted to
     /// the cache or evicted from the cache (so it should not have LRU nodes).
     is_admitted: AtomicBool,
-    /// `is_dirty` indicates that the entry has been inserted (or updated)
-    /// in the hash table, but the history of the insertion has not yet
-    /// been applied to the LRU deques and LFU estimator.
-    is_dirty: AtomicBool,
+    /// `entry_gen` (entry generation) is incremented every time the entry is updated
+    /// in the concurrent hash table.
+    entry_gen: AtomicU16,
+    /// `policy_gen` (policy generation) is incremented every time entry's WriteOpe
+    /// is applied to the cache policies including the LRU deque and LFU estimator.
+    policy_gen: AtomicU16,
     last_accessed: AtomicInstant,
     last_modified: AtomicInstant,
     expiration_time: AtomicInstant,
@@ -29,7 +31,8 @@ impl<K> EntryInfo<K> {
         Self {
             key_hash,
             is_admitted: AtomicBool::default(),
-            is_dirty: AtomicBool::new(true),
+            entry_gen: AtomicU16::new(0),
+            policy_gen: AtomicU16::new(0),
             last_accessed: AtomicInstant::new(timestamp),
             last_modified: AtomicInstant::new(timestamp),
             expiration_time: AtomicInstant::default(),
@@ -54,12 +57,40 @@ impl<K> EntryInfo<K> {
 
     #[inline]
     pub(crate) fn is_dirty(&self) -> bool {
-        self.is_dirty.load(Ordering::Acquire)
+        self.entry_gen.load(Ordering::Acquire) != self.policy_gen.load(Ordering::Acquire)
     }
 
     #[inline]
-    pub(crate) fn set_dirty(&self, value: bool) {
-        self.is_dirty.store(value, Ordering::Release);
+    pub(crate) fn entry_gen(&self) -> u16 {
+        self.entry_gen.load(Ordering::Acquire)
+    }
+
+    /// Increments the entry generation and returns the new value.
+    #[inline]
+    pub(crate) fn incr_entry_gen(&self) -> u16 {
+        // NOTE: This operation wraps around on overflow.
+        let prev = self.entry_gen.fetch_add(1, Ordering::AcqRel);
+        prev.wrapping_add(1)
+    }
+
+    /// Sets the policy generation to the given value.
+    #[inline]
+    pub(crate) fn set_policy_gen(&self, value: u16) {
+        let g = &self.policy_gen;
+        loop {
+            let current = g.load(Ordering::Acquire);
+            // TODO: Find a way to handle the case that the current value has been
+            // wrapped around. In such a case, `current < value`` actually means
+            // `current > value`.
+            if current >= value {
+                break;
+            }
+            if g.compare_exchange_weak(current, value, Ordering::AcqRel, Ordering::Acquire)
+                .is_ok()
+            {
+                break;
+            }
+        }
     }
 
     #[inline]
--- a/src/common/concurrent/entry_info.rs
+++ b/src/common/concurrent/entry_info.rs
@@ -6,14 +6,14 @@ use crate::common::{concurrent::atomic_time::AtomicInstant, time::Instant};
 #[derive(Debug)]
 pub(crate) struct EntryInfo<K> {
     key_hash: KeyHash<K>,
-    /// `is_admitted` indicates that the entry has been admitted to the
-    /// cache. When `false`, it means the entry is _temporary_ admitted to
-    /// the cache or evicted from the cache (so it should not have LRU nodes).
+    /// `is_admitted` indicates that the entry has been admitted to the cache. When
+    /// `false`, it means the entry is _temporary_ admitted to the cache or evicted
+    /// from the cache (so it should not have LRU nodes).
     is_admitted: AtomicBool,
     /// `entry_gen` (entry generation) is incremented every time the entry is updated
     /// in the concurrent hash table.
     entry_gen: AtomicU16,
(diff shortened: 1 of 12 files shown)


src/future/cache.rs
```diff
@@ -3062,6 +3062,44 @@ mod tests {
...
+        cache.insert("b", "bob").await;
+        cache.insert("c", "cathy").await; // c1
+        mock.increment(Duration::from_secs(2));
+
+        // The following `insert` will do the followings:
+        // 1. Replaces current "c" (c1) in the concurrent hash table (cht).
+        // 2. Runs the pending tasks implicitly.
+        //    (1) c1 will be evicted by size constraint.
+        //    (2) "a" will be admitted.
+        //    (3) "b" will be admitted.
```
tatsuya6502, line 3085:
```suggestion
        //    (1) "a" will be admitted.
        //    (2) "b" will be admitted.
        //    (3) c1 will be evicted by size constraint.
```

tatsuya6502: Applied the change:
https://github.com/moka-rs/moka/pull/348/commits/a28970ca7f1ab68a7adaa802b3739efb1e3f5212#diff-338f772312880199137f8bbb508f15dd7f83457565cc983b85043cddd71b2f24

tatsuya6502:
To clarify the remaining tasks, added the tasks section with check boxes to [the description](https://github.com/moka-rs/moka/pull/348#issue-2000689436) of this PR.

tatsuya6502:
@peter-scholtens 

FYI, I will be unable to work on Moka and other products for a couple of weeks from now. I will be in a hospital for medical treatments. I will have some chances to check GitHub Issues and Discussions from there, but will not actively respond to them.

Sorry for the inconvenience. I will resume work on this PR when I return from the hospital and everything is settled.

peter-scholtens:
Take your time, health is the most important thing in life.

tatsuya6502:
Updated the [description](https://github.com/moka-rs/moka/pull/348#issue-2000689436) of this issue.

src/common/concurrent/entry_info.rs
```diff
@@ -1,19 +1,22 @@
...
+    /// from the cache (so it should not have LRU nodes).
     is_admitted: AtomicBool,
-    /// `is_dirty` indicates that the entry has been inserted (or updated)
-    /// in the hash table, but the history of the insertion has not yet
-    /// been applied to the LRU deques and LFU estimator.
-    is_dirty: AtomicBool,
+    /// `entry_gen` (entry generation) is incremented every time the entry is updated
+    /// in the concurrent hash table.
+    entry_gen: AtomicU16,
+    /// `policy_gen` (policy generation) is incremented every time entry's `WriteOpe`
```
tatsuya6502, line 16:
```suggestion
    /// `policy_gen` (policy generation) is incremented every time entry's `WriteOp`
```

Will fix in a separate PR #363.

tatsuya6502: Fixed the typo by this commit for this PR:

- https://github.com/moka-rs/moka/pull/348/commits/b866dce0e9343d17bb808c9b35c87503fbbe28c0#diff-96c4a5be5f17869907f65e314ff6a08126fb0aef7489359b9ebdfcf2984ee399R16

tatsuya6502 (commented):
Merging.

Swatinem:
FYI, I believe this fixed a long standing issue we were observing where the cache hit rate would slowly degrade over time after ~1 week of operation.

Updating to `0.12.2`, the hit rate has been stable without any degradation for 2+ weeks:

image

tatsuya6502:
Thank you for reporting. That is a great news!

I didn't realize it until you mentioned it, but they should be related. The entries affected by this bug are counted in the cache usage even though they have no value (thus cache misses). Since they will never be evicted, the hit rate gradually decreases as the cache is occupied by these broken entries.

PR #355, opened 2023-12-11
Fix the CI with minimal crate versions. (moka v0.11.x)

tatsuya6502:
Change the following dev dependencies:

- Raise the minimal `async-std` version from 1.1 to 1.2.
- Use `rustls` in `reqwest` instead of `native-tls` (openssl on Linux).


tatsuya6502 (commented):
Merging.

PR #357, opened 2023-12-12
 Fix the CI with the MSRV 1.60.0 and also nightly + minimal crate versions. (moka v0.9.x) 

tatsuya6502:
Change the following _dev_ dependencies for the MSRV 1.60.0:

- Pin `async-global-executor` version to v2.3.1.
- Pin `blocking` version to v1.4.1.

Change the following _dev_ dependency:

- Pin `proc-macro2` to v1.0.60.


tatsuya6502 (commented):
Merging.

PR #362, opened 2023-12-23
 Fix the CI with minimal crate versions. (moka v0.11.x)

tatsuya6502:
- Change the following dev dependencies:
    - Change proc-macro2 from v1.0.60 to v1.0.63.
- Disable fail-fast feature on CI and CIQuantaDisabled workflows


tatsuya6502 (commented):
Merging.

PR #363, opened 2023-12-24
Update the access time of a cached entry immediately after the entry is read

### LINKED ISSUE #359: [BUG] time_to_idle not working as expected
`time_to_idle` function giving unexpected results. Follow below steps to replicate the issue.

```
[dependencies]
moka = { version = "0.12.1", features = ["sync"] }
```

```
fn main() {
        .max_capacity(10)
        .time_to_idle(std::time::Duration::from_secs(5))
        .build();

    cache.insert(1, 1);

    std::thread::sleep(std::time::Duration::from_secs(4));
    println!("{:?}", cache.get(&1));

    std::thread::sleep(std::time::Duration::from_secs(2));
    println!("{:?}", cache.get(&1));
    println!("{:?}", cache.get(&1));
}

```

**Output**
```
Some(1)
None
Some(1)

```
**Expected Output**
```
Some(1)
Some(1)
Some(1)
```

tatsuya6502:
Fixes #359.

~~(We will merge this into the `main` branch after #348 is merged)~~

## Problem

When a cached entry is read (by e.g. the `get` method) or write (by e.g. the `insert` method), the `last_accessed` time of the entry is updated. When `time_to_idle` of the `Cache` is set, the `last_accessed` time is used to determine whether the entry is expired or not.

Up to Moka `v0.12.1`, `last_accessed` time was _eventual consistent_ for read operations. So it will not be updated immediately after read, but will be updated sometime later when pending tasks are processed. (Note: It will be updated to the time when read was performed, not to the time when pending task is processed). This was designed so to achieve better throughput as updating `last_accessed` has a small performance overhead from a synchronous primitive such as `AtomicU64` when the default features are enabled, and `Mutex` when the `atomic64` or `quanta` feature is disabled. 

Note that `last_accessed` time is _strong consistent_ for write operations, and it is updated during the write.

Before `v0.12.0`, pending tasks were processed periodically (every ~0.3 secs) by a background thread, so the delay after read was not obvious. However we removed the background thread completely from `v0.12.0`, and the delay becomes obvious.

## Fix

Like the write operations, make `last_accessed` strong consistent for read operations.

### DIFF
--- a/src/common/concurrent.rs
+++ b/src/common/concurrent.rs
@@ -303,7 +303,6 @@ impl<K, V> AccessTime for TrioArc<ValueEntry<K, V>> {
 pub(crate) enum ReadOp<K, V> {
     Hit {
         value_entry: TrioArc<ValueEntry<K, V>>,
-        timestamp: Instant,
         is_expiry_modified: bool,
     },
     // u64 is the hash of the key.
--- a/src/future/base_cache.rs
+++ b/src/future/base_cache.rs
@@ -323,12 +323,13 @@ where
                         );
                     }
 
+                    entry.set_last_accessed(now);
+
                     let maybe_key = if need_key { Some(Arc::clone(k)) } else { None };
                     let ent = Entry::new(maybe_key, entry.value.clone(), false);
                     let maybe_op = if record_read {
                         Some(ReadOp::Hit {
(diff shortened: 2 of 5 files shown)


tatsuya6502 (commented):
Merging.

PR #365, opened 2023-12-28
Reduce the TTL from 3 secs to 2.5 secs in the cascading drop example

tatsuya6502:
As we discussed in https://github.com/moka-rs/moka/pull/350#issuecomment-1870995237, we will reduce the TTL value in the example from 3 secs to 2.5 secs.

Also this fixes a Clippy warning from a recent beta Clippy `0.1.75 (1a06ac5b5d7 2023-12-01)`.

* * *
CC: @peter-scholtens

### DIFF
--- a/examples/cascading_drop_async.rs
+++ b/examples/cascading_drop_async.rs
@@ -115,11 +115,17 @@ async fn main() {
         }
     });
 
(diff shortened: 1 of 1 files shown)


tatsuya6502 (commented):
Merging.

PR #375, opened 2024-01-09
Compute API - Update the tests to ensure the internal waiter map is empty at the end

tatsuya6502:
Relates to #329, a bug in `v0.12.0` that was fixed for `v0.12.1`.

Update the tests for `get_with`, `and_compute_with`, etc. to ensure the internal waiter map is empty at the end.


tatsuya6502 (commented):
Merging.

PR #376, opened 2024-01-09
Raise the minimum version of the `quanta` to `v0.12.2`

tatsuya6502:
Raise the version requirement of the `quanta` from `>=0.11.0, <0.12.0` to `>=0.12.2, <0.13.0` to avoid under-measuring the elapsed time on Apple silicon Macs.

- Due to this under-measurement, cached entries expire sightly later than expected on macOS arm64.
- For more details, see https://github.com/metrics-rs/quanta/issues/96.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@
         - `and_try_compute_with` method, which is similar to above but returns
           `Result`.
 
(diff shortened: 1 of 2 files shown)


tatsuya6502 (commented):
Merging.

PR #382, opened 2024-01-19
Add an example called `reinsert_expired_entries_sync`

### LINKED ISSUE #298: How can i add something to the cache inside the `eviction_listener`?
Hey Moka team, i'm having a special scenario that i want to re add an item or a modified version of it to the cache when the item is getting evicted in certain scenarios, how can I achieve that?

tatsuya6502:
Add an example to demonstrate how to write an eviction listener that will reinsert the expired entries.
Fixes #298.

Also bump the version to `v0.12.4`.


tatsuya6502 (commented):
Merging.

PR #384, opened 2024-01-20
Ensure `crossbeam-epoch` to run GC when dropping a cache

tatsuya6502:
Fixes #383.

This PR ensures `crossbeam-epoch` to run GC (run the `deferred_fn`s in the global bag) when dropping a `moka` cache. This should help the entries, who are currently or previously in the cache, to be dropped.

We have been struggling this `crossbeam-epoch` spec for years. I added a comment to [our notes about this issue](https://github.com/orgs/moka-rs/projects/1?pane=issue&itemId=1314133).

Issue #383, opened 2024-01-19
entries are leaked when sync::Cache is dropped


igorzi:
Simple repro below (`Foo` is not dropped when the cache is dropped).

```rust
use moka::sync::Cache;

struct Foo {}

impl Drop for Foo {
  fn drop(&mut self) {
    eprintln!("Foo dropped");
  }
}

fn main() {
  let foo = Arc::new(Foo {});
  let cache = Cache::builder().build();
  {
    cache.get_with("foo", move || {
      foo
    });
  }
  drop(cache);
}
```

igorzi:
Note that if I add `cache.run_pending_tasks();` right before `drop(cache);` - the entries do get dropped.

tatsuya6502:
Thanks for reporting!

I believe this is due to the spec that `crossbeam-epoch` crate does not guarantee to run the destructor of the pointed object.

https://docs.rs/crossbeam-epoch/0.9.18/crossbeam_epoch/struct.Guard.html#method.defer_destroy

> There is no guarantee when exactly the destructor will be executed. The only guarantee is that it won’t be executed until all currently pinned threads get unpinned. In theory, the destructor _might never run_, but the epoch-based garbage collection will make an effort to execute it reasonably soon.

We have been struggling with this spec ([notes](https://github.com/orgs/moka-rs/projects/1?pane=issue&itemId=1314133)) and added some mitigations. Now I am adding another mitigation via #384, but the new unit tests are a bit flaky.

![gh-actions](https://github.com/moka-rs/moka/assets/159403/6819a334-68fd-40d5-bdf0-94a72ec39e31)

Let me struggle a bit longer to see if I can stabilize them.

tatsuya6502:
I am not totally sure if I have stabilized the flaky tests, but I am going to merge #384 and will see how these jobs will go. If everything goes well, I will publish a new version `moka@v0.12.4` to crates.io. Hopefully that will happen in the next few days.

Also I created the following issue to evaluate an alternatives of `crossbeam-epoch`:

- #385

Thanks again for reporting the issue!

igorzi:
Thanks!

tatsuya6502:
FYI, I just published [`moka@v0.12.4`](https://crates.io/crates/moka/0.12.4) to crates.io.

PR #387, opened 2024-01-21
CI: Stabilize `ensure_gc_runs_when_dropping_cache` test for `sync::Cache`

tatsuya6502:
Run a flaky test `ensure_gc_runs_when_dropping_cache` for `sync::Cache` separately from other tests to avoid interference from other tests running in parallel.

The test was added by #384 to mitigate #383.


tatsuya6502 (commented):
Merging.

Issue #414, opened 2024-04-10
Miri errors on the `timer_wheel` module. (Miri v0.1.0 8b2459c1 2024-04-09)


tatsuya6502:
Miri started to report errors on the test for the `moka::common::timer_wheel` module.

https://github.com/moka-rs/moka/actions/runs/8629705772/job/23654526179

```console
$ cargo +nightly miri -V
miri 0.1.0 (8b2459c1 2024-04-09)
```

The errors change when [Stacked Borrows](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md) is used, or [Tree Borrows]() is used.

## Miri's Stacked Borrow

```console
$ cargo +nightly miri test timer_wheel --features 'sync, future'
```

<details>

```console
running 2 tests
test common::timer_wheel::tests::test_advance ... error: Undefined Behavior: trying to retag from <704136> for SharedReadOnly permission at alloc237477[0x0], but that tag does not exist in the borrow stack for this location
   --> /Users/tatsuya/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/non_null.rs:402:18
    |
402 |         unsafe { &*self.as_ptr().cast_const() }
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                  |
    |                  trying to retag from <704136> for SharedReadOnly permission at alloc237477[0x0], but that tag does not exist in the borrow stack for this location
    |                  this error occurs as part of retag at alloc237477[0x0..0x28]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <704136> was created by a Unique retag at offsets [0x0..0x28]
   --> src/common/deque.rs:172:37
    |
172 |             let node = NonNull::new(Box::into_raw(node)).expect("Got a null ptr");
    |                                     ^^^^^^^^^^^^^^^^^^^
help: <704136> was later invalidated at offsets [0x0..0x28] by a SharedReadOnly retag
   --> src/common/timer_wheel.rs:517:55
    |
517 | ...                   let entry_info = unsafe { node_p.as_ref() }.element.entry_info();
    |                                                 ^^^^^^^^^^^^^^^
[...]

PR #415, opened 2024-04-10
Prevent pending `run_pending_tasks` of `future::Cache` from causing busy loop in `schedule_write_op` 

### LINKED ISSUE #412: Pending `run_pending_tasks` may cause busy loop in `schedule_write_op`
It's not so easy to create a minimal reproducible example. I will try to describe my code and my findings.

As a workaround for #408, I run `run_pending_tasks` manually when the cache size exceeds the capacity by a wide margin. However, occasionally I find my tokio worker threads are stuck in the busy loop of `schedule_write_op`.

I examine what happens with gdb:

- The write op channel is full. I'm doing a lot of cache insertions so it's not strange.
- All worker threads cannot acquire the `current_task` mutex in `try_run_pending_tasks`.
- The internal state of the `current_task` mutex is `2`, which means the mutex is not locked but there is a starved acquirer. In this state, `try_lock` will fail.

I further add some logs to moka and find that inside my manual call to `run_pending_tasks`, it tries to acquire the `current_task` mutex and does not actually get it acquired. It probably shows that the starved mutex acquirer is the manual `run_pending_tasks` task.

There are various reasons why the `run_pending_tasks` task is not waked up to acquire the lock:

- **All** of the tokio worker threads are running the busy loops in `schedule_write_op`. The runtime never schedules the `run_pending_tasks` task to run.
- I also combine the `run_pending_tasks` task together with other `insert` tasks in a `join_all` future. So, when the `join_all` future is doing the busy loop, it's also impossible for the `run_pending_tasks` task to run.

tatsuya6502:
Fixes #412.

## Summary

This PR prevents the async runtime's schedulers from going into infinite busy loops in an internal `schedule_write_op` method when there are pending `run_pending_tasks` calls on the `future::Cache`.

- This bug was introduced in `v0.12.0` when the background threads were removed from `future::Cache`.
- This bug can occur when `run_pending_task` method is called by user code while cache is receiving a high number of concurrent cache write operations such as `insert`, `get_with` or `invalidate`.
- When it occurs, the `schedule_write_op` method will be spinning in a busy loop forever, causing high CPU usage and  all other async tasks to be starved..

The fix is to replace a `async_lock::RwLock` used in `schedule_write_op` method with an event notification mechanism provided by the `event-listener` crate. (Note that `event-listener` is used by `async_lock::RwLock` too.)

### Other Changes

This PR also does the followings:

- Bump the `moka` version to `v0.12.6`.
- Upgrades `async-lock` crate used by `future::Cache` to the latest version 3.3. (This is not related to the bug)

## The Root Cause

When a cache write operation such as `insert` is performed on `future::Cache`, `schedule_write_op` method is called internally to send a write operation log to a channel. Later, an internal `do_run_pending_tasks` method will be called and it will receive the log from the channel. The channel is bounded so it can only hold up to a fixed number of items (logs). When the channel gets full, `schedule_write_op` method will fail to send a log to the channel, so it will retry to send until it succeeds.

`future::Cache` used `async_lock::RwLock` to prevent the retry loop to take all CPU time. It will wait for a read lock becomes available, so that the scheduler threads of the async runtime can do other stuff including running `do_run_pending_tasks`. `do_run_pending_tasks` acquires an exclusive write lock at the beginning, so the retry loop will be kept waiting while `do_run_pending_tasks` is running.

The problem with the current implementation is the retry loop will keep spinning when `do_run_pending_tasks` is not running because no write lock is taken. So the retry loop can occupy the scheduler thread and never yield to other async tasks until `do_run_pending_tasks` is started. If there are enough number of retry loop spinning in parallel, all the scheduler threads will be occupied by them, preventing `do_run_pending_tasks` to start! This causes the retry loop to keep spinning forever by occupied schedulers.

## The Fix

This PR fixes the problem by replacing the `RwLock` with an event notification mechanism provided by the `event-listener` crate. The retry loop in `schedule_write_op` will wait for an event to arrive no matter if `do_run_pending_tasks` is running. This ensures async schedulers not to be occupied by the retry loop in `schedule_write_op`, so the schedulers can run `do_run_pending_tasks` whenever needed.

The event will be sent when one of the following conditions meet:

- The `do_run_pending_tasks` method has removed some logs from the channel.
- The `Housekeeper`'s `run_pending_tasks` or `try_run_pending_tasks` method has freed a lock on a  `Mutex` called `current_task`.
    - `current_task` is used to ensure only one `run_pending_tasks` or `try_run_pending_tasks` method to run at a time.
    - `run_pending_tasks` and `try_run_pending_tasks` call `do_run_pending_tasks`.

The latter is needed to make the retry loop to spin once more. This may start `do_run_pending_tasks` again because the retry loop will call `try_run_pending_tasks` if necessary before resending the log to the channel.

NOTE: You may wonder why the spinning retry loop cannot start `do_run_pending_tasks` if the loop itself is calling `try_run_pending_tasks`. This is because if there are any `run_pending_task` calls waiting for `current_task` lock to be available, one of them will be given the right to acquire the lock, making no other to do so. Unless an async scheduler runs exactly the one having the right to acquire, `do_run_pending_tasks` will not start. See the _starved lock_ in [the description of #412](https://github.com/moka-rs/moka/issues/412#issue-2232606128) for the details.

## Tests

- Tested by the program attached to https://github.com/moka-rs/moka/issues/412#issuecomment-2047896594 and verified that the problem was not reproduced.
- Tested by running [Mokabench](https://github.com/moka-rs/mokabench/) with `cargo run --release` and it completed without any new issues.
    - NOTE: Mokabench does not call `run_pending_tasks` so it will not reproduce the problem. The purpose of this test was to find any regressions.

### DIFF
--- a/src/future/base_cache.rs
+++ b/src/future/base_cache.rs
@@ -1415,10 +1415,7 @@ where
     /// Notifies all the async tasks waiting in `BaseCache::schedule_write_op` method
     /// for the write op channel to have enough room.
     fn notify_write_op_ch_is_ready(&self) {
-        let listeners = self.write_op_ch_ready_event.total_listeners();
-        // NOTE: The `notify` method accepts 0, so no need to check if `listeners` is
-        // greater than 0.
-        self.write_op_ch_ready_event.notify(listeners);
+        self.write_op_ch_ready_event.notify(usize::MAX);
     }
 
     fn now(&self) -> Instant {
--- a/src/future/base_cache.rs
+++ b/src/future/base_cache.rs
@@ -677,8 +677,8 @@ where
                 //
                 // - The `Inner::do_run_pending_tasks` method has removed some ops
                 //   from the write op channel.
-                // - THe `Housekeeper`'s `run_pending_tasks` or `
-                //   try_,run_pending_tasks` methods has freed the lock on the
+                // - The `Housekeeper`'s `run_pending_tasks` or `
+                //   try_run_pending_tasks` methods has freed the lock on the
                 //   `current_task`.
                 //
                 ch_ready_event.listen().await;
--- a/src/future/base_cache.rs
+++ b/src/future/base_cache.rs
@@ -121,8 +121,8 @@ impl<K, V, S> BaseCache<K, V, S> {
     }
 
     #[inline]
-    pub(crate) fn maintenance_task_lock(&self) -> &RwLock<()> {
-        &self.inner.maintenance_task_lock
+    pub(crate) fn write_op_ch_ready_event(&self) -> &event_listener::Event<()> {
+        &self.inner.write_op_ch_ready_event
     }
 
     pub(crate) fn notify_invalidate(
@@ -618,7 +618,7 @@ where
     pub(crate) async fn schedule_write_op(
         inner: &Arc<impl InnerSync + Send + Sync + 'static>,
         ch: &Sender<WriteOp<K, V>>,
-        maintenance_task_lock: &RwLock<()>,
+        ch_ready_event: &event_listener::Event<()>,
         op: WriteOp<K, V>,
         ts: Instant,
         housekeeper: Option<&HouseKeeperArc>,
@@ -638,6 +638,7 @@ where
         let mut op = op;
         let mut spin_loop_attempts = 0u8;
         loop {
+            // Run the `Inner::do_run_pending_tasks` method if needed.
             BaseCache::<K, V, S>::apply_reads_writes_if_needed(
                 Arc::clone(inner),
                 ch,
@@ -645,6 +646,8 @@ where
                 housekeeper,
             )
             .await;
+
+            // Try to send our op to the write op channel.
             match ch.try_send(op) {
                 Ok(()) => return Ok(()),
                 Err(TrySendError::Full(op1)) => {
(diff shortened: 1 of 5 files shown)


tatsuya6502:
The CI for Miri tests are failing. It is not related to this PR and we will track it by #414.

src/future/base_cache.rs
```diff
@@ -666,25 +669,22 @@ where
...
+                // Yield the async runtime scheduler to other async tasks and wait
+                // for a channel ready event. This event will be sent when one of the
+                // following conditions is met:
                 //
-                // `do_run_pending_tasks` will receive some of the ops from the
-                // channel and apply them to the data structures for the cache
-                // policies, so the channel will have some room for the new ops.
+                // - The `Inner::do_run_pending_tasks` method has removed some ops
+                //   from the write op channel.
+                // - THe `Housekeeper`'s `run_pending_tasks` or `
```
tatsuya6502, line 680:
```suggestion
                // - The `Housekeeper`'s `run_pending_tasks` or `
```

src/future/base_cache.rs
```diff
@@ -666,25 +669,22 @@ where
...
+                //   try_,run_pending_tasks` methods has freed the lock on the
+                //   `current_task`.
```
tatsuya6502, line 682:
```suggestion
                // - The `Housekeeper`'s `run_pending_tasks` or `
                //   try_run_pending_tasks` methods has freed the lock on the
                //   `current_task`.
```

tatsuya6502 (commented):
Merging.

PR #417, opened 2024-04-14
Make a single call to `run_pending_tasks` to evict as many entries as possible from the cache

### LINKED ISSUE #408: Inefficient size aware eviction when inserting new entries with larger size
Example code (LRU policy, moka 0.12.5):

```rust
use moka::{future::CacheBuilder, policy::EvictionPolicy};
use rand::prelude::*;

#[tokio::main]
async fn main() {
    let cache = CacheBuilder::<u64, Vec<u8>, _>::new(1 << 24)
        .eviction_policy(EvictionPolicy::lru())
        .weigher(|_, v| v.len() as u32)
        .build();

    // Fill the cache with 512 bytes values
    for i in 0..65536 {
        let mut value = vec![0; 512];
        thread_rng().fill_bytes(&mut value);
        cache.insert(i, value).await;
    }
    println!("{}", cache.weighted_size());

    // Insert much larger values
    for i in 0..4096 {
        let mut value = vec![0; 16384];
        thread_rng().fill_bytes(&mut value);
        cache.insert(65536 + i, value).await;
    }
    println!("{}", cache.weighted_size());

    while cache.weighted_size() > 1 << 24 {
        println!("{} {}", cache.entry_count(), cache.weighted_size());
    }
}
```

Output:

```
16777216
66709504
5300 66709504
4864 67502080
4364 67246080
3864 63307776
3364 55115776
2864 46923776
2364 38731776
1864 30539776
1364 22347776
```

After inserting 4096 16k-size values, the total weighted size increases and exceeds the 16M capacity.

tatsuya6502:
Fixes #408.

## Changes

This PR ensures that a single call to `run_pending_tasks` to evict as many entries as possible from the cache.

- Disable the batch limits when the eviction listener is not set to the cache.
  - So single `run_pending_tasks` call should remove all the entries that can be evicted from the cache.
- Add a hard-coded maintenance task timeout duration.
  - When the eviction listener is set, `run_pending_tasks` should stop (return) after the timeout duration is elapsed.
  - This is a safe-guard to prevent the maintenance task from running long time when the user provided a slow eviction listener.
  - In older versions, the batch size was used to limit the time spent on the maintenance task.
  - Note that `run_pending_tasks` checks the timeout only after processing a batch of entries, so `run_pending_tasks` can run longer than the timeout duration.
- Reduce the size of the read op and write op channels.

## Tests

1. Add a unit test using a cache with _no_ eviction listener. It verifies that a single call to `run_pending_tasks` to evict as many entries as possible.
2. Add a unit test using a cache with a eviction listener. It verifies that timeout duration is working.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Changed
 
 - Ensure a single call to `run_pending_tasks` to evict as many entries as possible
-  from the cache ([#417](gh-pull-0417).
+  from the cache ([#417][gh-pull-0417]).
 
 
 ## Version 0.12.6
@@ -880,6 +880,7 @@ The minimum supported Rust version (MSRV) is now 1.51.0 (Mar 25, 2021).
 [gh-issue-0034]: https://github.com/moka-rs/moka/issues/34/
 [gh-issue-0031]: https://github.com/moka-rs/moka/issues/31/
 
+[gh-pull-0417]: https://github.com/moka-rs/moka/pull/417/
 [gh-pull-0390]: https://github.com/moka-rs/moka/pull/390/
 [gh-pull-0384]: https://github.com/moka-rs/moka/pull/384/
 [gh-pull-0382]: https://github.com/moka-rs/moka/pull/382/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Moka Cache &mdash; Change Log
 
+## Version 0.12.7
+
+### Changed
+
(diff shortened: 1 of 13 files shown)


tatsuya6502:
A CI job for Kani verifier is failing with compile errors on `procmacro2` crate. It is not related to the changes in this pull request.

We will address the Kani and `procmacro2` issue separately in the following: 

- https://github.com/moka-rs/moka/issues/418

tatsuya6502 (commented):
Merging.

PR #421, opened 2024-05-03
docs: fix per-entry expiration policy documentation

arcstur:
* remove unused imports
* add feature time to tokio, since the example uses tokio::time
* do cache.run_pending_tasks() before emitting the entry count so that it's accurate


tatsuya6502 (approved):
Thank you! It looks good to me.

PR #434, opened 2024-07-06
Fix the CI. 2024-07-06 (v0.12.x)

tatsuya6502:
- Fix the CI for the MSRV 1.65.
    - Pin `actix-rt` to v2.9.0 for the MSRV 1.65, otherwise the build will fail.
- Fix the CI for Rust 1.80 beta
    - Make `rustc` to allow specific custom `cfg` condition names (e.g., `rustver`).


tatsuya6502 (commented):
Merging.

PR #436, opened 2024-07-07
Remove an experimental `js` feature

### LINKED ISSUE #423: Should Moka work in WASM?
There is a `js` feature which enables `uuid/js` feature, to use a web safe random generator. However, it seems that Moka still uses `time` somewhere, because in the browser the app panics on time not implemented on this platform. I would love to be able to use Moka in a web app (rust wasm), but am not sure if this is a use case that is intended to be supported, and currently broken (or I am doing something wrong), or if this is simply not what Moka is meant for.

Thanks!

tatsuya6502:
Remove the `js` feature.

The `js` feature is experimental and it is actually broken (will not compile). Moka currently does not support Wasm and WASI targets.

Fixes #423.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,11 +35,6 @@ logging = ["log"]
 # https://github.com/moka-rs/moka#resolving-compile-errors-on-some-32-bit-platforms
 atomic64 = []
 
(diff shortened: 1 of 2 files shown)


tatsuya6502 (commented):
Merging.

PR #437, opened 2024-07-07
CI: Run Cirrus CI only by cron

tatsuya6502:
We use Cirrus CI for testing Moka on real Arm64 machine. However, it takes very long time (about 25 minutes) to complete.

Change Cirrus CI job to be manually triggered, and set up a cron schedule for it. (I hope a manually triggered job can be automatically triggered by cron, but it is not clear on [the doc](https://cirrus-ci.org/guide/writing-tasks/#cron-builds))

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -4,6 +4,12 @@ env:
   NUM_CPUS: "2"
 
 linux_arm64_task:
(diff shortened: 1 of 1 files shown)


tatsuya6502 (commented):
Merging.

PR #445, opened 2024-07-21
Remove needless traits along with `async-trait` usage

Swatinem:
I originally just found a redundant `Arc::clone` within `GetOrInsert::insert`.
But then I discovered that `GetOrInsert` only has a single implementor and is thus unnecessary indirection.

The similar thing was true for `GetOrRemoveEntry` and `InnerSync`, so I removed those as well. One reason they might have existed was using a trait for code encapsulation, as removing them resulted in a bit more `pub(crate)` visibility, and a bunch more trait bounds all over the place.

However, in particular `#[async_trait]` usage was allocating `Box<dyn Future>` and using dynamic dispatch at runtime. Now that overhead is avoided and the compiler should be able to produce better code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ default = ["atomic64", "quanta"]
 sync = []
 
 # Enable this feature to use `moka::future::Cache`.
-future = ["async-lock", "async-trait", "event-listener", "futures-util"]
+future = ["async-lock", "event-listener", "futures-util"]
 
 # Enable this feature to activate optional logging from caches.
 # Currently cache will emit log only when it encounters a panic in user provided
@@ -60,7 +60,6 @@ quanta = { version = "0.12.2", optional = true }
 
 # Optional dependencies (future)
 async-lock = { version = "3.3", optional = true }
-async-trait = { version = "0.1.58", optional = true }
 event-listener = { version = "5.3", optional = true }
 futures-util = { version = "0.3.17", optional = true }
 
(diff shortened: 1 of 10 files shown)


tatsuya6502 (approved):
Thank you very much for working on this change! It looks good to me.

For just in case you are still curious why those traits existed, here is a bit of the history:

`GetOrRemoveEntry` and `InnerSync` traits were needed by `moka` prior to v0.12 to meet the trait and lifetime bounds that `ScheduledThreadPool` (from `scheduled-thead-pool` crate) required. I do not remember the very details, but it was hard to give a closure having an `Arc<base_cache::Inner>` to `ScheduledThreadPool` as `Inner` had many internal data structures including ones who are `!Send`. So I created these traits to hide (encapsulate) the internal data structures.

When I worked for `moka` v0.12 and removed `ScheduledThreadPool`, I tried to minimize the changes, so I did not touch the traits and their implementations.

I added `GetOrInsert` trait when I worked on `moka` v0.12. Some of the `ValueInitializer` methods used to take closures to get or insert an cache entry. When I had to create an async version of `ValueInitializer` for v0.12, I simply translated the closures to trait methods as Rust did not support async closures. (Later, I added the `remove` method to the trait when I added `entry().and_compute()` method.)

But, `GetOrInsert` trait was not absolutely necessary and now I think I should have tried what you did; to pass a reference to `base_cache::Inner`.

PR #452, opened 2024-08-23
Stop creating references unnecessarily to compare pointers by-address

JoJoDeveloping:
The test `common::deque::basics` contains several lines that look like this:
```
assert!(std::ptr::eq(head_d, unsafe { node1_ptr.as_ref() }));
```

What is happening is that we assert that `head_c`, a reference, is equal by-address to `node1_ptr`, a `NonNull<_>`. However, the way we do this is by creating a reference, instead of using `as_ptr`.

The current way has several downsides:
* It requires an `unsafe` block. Comparing pointer addresses is safe so this should be unnecessary.
* **It is unsound**, at least under [Tree Borrows](https://www.ralfj.de/blog/2023/06/02/tree-borrows.html) rules. The references created there violate the uniqueness of references/Boxes created internally in the `Deque`, which trips the `Deque` later when it tries to use these internally-held references.
* It is unnecessary -- comparing the result of `as_ptr` works just the same, maybe even faster due to less intermediary methods.

As such, the test should be updated to only use `as_ptr`. When done, it actually makes the tests pass under Tree Borrows.

See also https://github.com/moka-rs/mini-moka/pull/31

### DIFF
--- a/src/common/deque.rs
+++ b/src/common/deque.rs
@@ -373,7 +373,7 @@ mod tests {
         assert!(deque.contains(head_b));
         assert!(deque.is_head(head_b));
         assert!(deque.is_tail(head_b));
-        assert!(std::ptr::eq(head_b, unsafe { node1_ptr.as_ref() }));
+        assert!(std::ptr::eq(head_b, node1_ptr.as_ptr()));
         assert!(head_b.prev.is_none());
         assert!(head_b.next.is_none());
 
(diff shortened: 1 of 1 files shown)


tatsuya6502 (approved):
Thank you for submitting this pull request with detailed explanations! That is very very helpful.

Actually, we are already running CI with `miri` on the `deque` and `timer_wheel` modules. However it did not detect the issue you reported as we only used stacked borrows model. I created another pull request #454 to use tree borrows in the CI in addition to stacked borrows.

JoJoDeveloping:
Thanks for merging, and nice that you use Tree Borrows in your CI :partying_face: But be warned that in miri, Tree Borrows currently comes with a performance impact, since there's a bit more to check. While I plan to optimize this further, I can't promise anything. Though if you notice any code that is significantly slower under TB than SB, reach out and I will see what I can do.

Besides, can you also merge the [sister PR](https://github.com/moka-rs/mini-moka/pull/31) for mini-moka?

PR #454, opened 2024-08-25
CI: Run `miri` using tree borrows in addition to the existing stacked borrows

tatsuya6502:
Our CI runs `miri` on the following internal modules, which contain a bunch of `unsafe` codes.

- `deque`
- `timer_wheel`

And we have only used `miri`'s stacked borrows model to detect pointer aliasing violations.

This pull request adds `miri` tests using tree borrows model so that our CI will detect the problems like the one reported by #452.

https://github.com/moka-rs/moka/actions/runs/10545853093/job/29216602204
miri-tree-borrows

### DIFF
--- a/.github/workflows/Miri.yml
+++ b/.github/workflows/Miri.yml
@@ -47,8 +47,18 @@ jobs:
 
       - run: cargo miri setup
 
-      - name: Run Miri test (deque)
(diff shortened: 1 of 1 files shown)


tatsuya6502 (commented):
Merging.

PR #455, opened 2024-09-01
Update QuickCache table in README.md

arthurprs:
The original algorithm approximated CLOCK-pro, but deviated over time. Now that S3-FIFO has been published, it's actually much closer to that.


tatsuya6502 (approved):
Thank you!

PR #458, opened 2024-09-14
Fix the CI: 2024-09-15 v0.12.x

tatsuya6502:
- Pin Kani verifier to `v0.54.0` to avoid out of memory errors with `v0.55.0`.
- Pin `tokio-util` version to `0.7.11` to fix the CI for the MSRV `1.65.0`.
- Move the `check-cfg` lint stuff from the build script to `Cargo.toml`.


tatsuya6502 (commented):
Merging.

Issue #433, opened 2024-06-29
Feature Request: Alternative compute API that skips if other thread/async task is already computing for the key


tatsuya6502:
Split from:

- #432

As of `moka@v0.12.7`, concurrent calls on `Entry`'s `and_compute_with`/`and_try_compute_with` for the same key are serialized by the cache. If caller `A` and `B` called `and_compute_with` for the same key at the same time, only one of the closures `f` from either `A` or `B` is evaluated first. And once finish, another closure is evaluated (https://github.com/moka-rs/moka/issues/432#issuecomment-2196801694). These closures are always called, but not at the same time.

There will be some use cases that do not need/want the latter closure to be evaluated, and return immediately instead of blocking.

I have not came up with a good short name for the alternative methods yet. But they could be like the followings:

```rust
use moka::future::Cache;

let result = cache
    .entry_by_ref(&key)
    .and_try_compute_if_nobody_else(|entry| async move { 
        ... 
    }).await?;
```

Should I try the builder pattern?

```rust
let result = cache
    .entry_builder_by_ref(&key)
    .try_compute_with(|entry| async move {
        ...
    })
    .skip_if_already_in_progress()
    .run()
    .await?;
```

xuehaonan27:
Is there any plan to add this feature into moka? I'm eager to contribute!

xuehaonan27:
> Is there any plan to add this feature into moka? I'm eager to contribute!

I am new to this project, so I want to take this issue as my starting point! Much thanks if you could give me some guidelines!

xuehaonan27:
I have submitted a [PR](https://github.com/moka-rs/moka/pull/460) #460  implement this feature.
I'm looking forward to your reply!
Test code:
```Rust
use std::sync::Arc;

use moka::{
    future::Cache,
    ops::compute::{CompResult, Op},
};

const N: usize = 100000;
const CONCURRENCY: usize = 100; // concurrent task nums

#[tokio::main]
async fn main() {
    /*
        Note: the final result should less than `N`.
        That's okay because we are testing [`and_try_compute_if_nobody_else`] which should
        cancel some computations, rather than testing whether the addition itself is thread-safe
        or not.
    */

    // Uncomment this line to test computation serialized by cache.
    // You can only see
    // ReplacedWith(Entry { key: "key1", value: <xxx>, is_fresh: true, is_old_value_replaced: true })
    // in stdout, which is as expected.
    computation_serialized().await;

    // Uncomment this line to test new feature.
[...]

PR #460, opened 2024-10-09
Support and_try_compute_if_nobody_else

xuehaonan27:
Issue: #433 
It will only evaluate one closure for a certain entry, and other closures will be canceled, returning `CompResult::Unchanged`.

Add `ValueInitializer::post_init_for_try_compute_with_if_nobody_else`. 
Add `Cache::try_compute_if_nobody_else_with_hash_and_fun`. 
Add `RefKeyEntrySelector::and_try_compute_if_nobody_else`.

### DIFF
--- a/src/future/cache.rs
+++ b/src/future/cache.rs
@@ -1879,6 +1879,23 @@ where
             .await
     }
 
+    pub(crate) async fn try_compute_if_nobody_else_with_hash_and_fun<F, Fut, E>(
+        &self,
+        key: Arc<K>,
+        hash: u64,
+        f: F,
+    ) -> Result<compute::CompResult<K, V>, E>
+    where
+        F: FnOnce(Option<Entry<K, V>>) -> Fut,
+        Fut: Future<Output = Result<compute::Op<V>, E>>,
+        E: Send + Sync + 'static,
+    {
+        let post_init = ValueInitializer::<K, V, S>::post_init_for_try_compute_with_if_nobody_else;
+        self.value_initializer
+            .try_compute_if_nobody_else(key, hash, self, f, post_init, true)
+            .await
+    }
+
     pub(crate) async fn upsert_with_hash_and_fun<F, Fut>(
         &self,
         key: Arc<K>,
--- a/src/future/entry_selector.rs
+++ b/src/future/entry_selector.rs
@@ -878,6 +878,21 @@ where
             .await
     }
 
+    pub async fn and_try_compute_if_nobody_else<F, Fut, E>(
+        self,
+        f: F,
+    ) -> Result<compute::CompResult<K, V>, E>
+    where
+        F: FnOnce(Option<Entry<K, V>>) -> Fut,
+        Fut: Future<Output = Result<compute::Op<V>, E>>,
+        E: Send + Sync + 'static,
+    {
+        let key = Arc::new(self.ref_key.to_owned());
+        self.cache
+            .try_compute_if_nobody_else_with_hash_and_fun(key, self.hash, f)
+            .await
+    }
+
     /// Performs an upsert of an [`Entry`] by using the given closure `f`. The word
     /// "upsert" here means "update" or "insert".
     ///
--- a/src/future/value_initializer.rs
+++ b/src/future/value_initializer.rs
@@ -398,6 +398,144 @@ where
         // The lock will be unlocked here.
     }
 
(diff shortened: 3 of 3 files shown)


xuehaonan27:
Test code:
```Rust
use std::sync::Arc;

use moka::{
    future::Cache,
    ops::compute::{CompResult, Op},
};

const N: usize = 100000;
const CONCURRENCY: usize = 100; // concurrent task nums

#[tokio::main]
async fn main() {
    /*
        Note: the final result should less than `N`.
        That's okay because we are testing [`and_try_compute_if_nobody_else`] which should
        cancel some computations, rather than testing whether the addition itself is thread-safe
        or not.
    */

    // Uncomment this line to test computation serialized by cache.
    // You can only see
    // ReplacedWith(Entry { key: "key1", value: <xxx>, is_fresh: true, is_old_value_replaced: true })
    // in stdout, which is as expected.
    computation_serialized().await;

    // Uncomment this line to test new feature.
    // You can see lines like:
    // Unchanged(Entry { key: "key1", value: <xxx>, is_fresh: false, is_old_value_replaced: false })
    // in stdout, which is as expected because there are multiple waiters manipulating the same
    // entry simutaneously.
    // However, I am not sure whether [`CompResult`] like `Unchanged` is proper for
    // this situation where the computation is cancelled because of another waiter
    // occupying the entry.
    // Should we add a new Enum to [`CompResult`] to represent this situation?
    computation_maybe_cancelled().await;
}

async fn computation_serialized() {
    // Increment a cached `u64` counter.
    async fn inclement_counter(cache: &Cache<String, u64>, key: &str) -> CompResult<String, u64> {
        cache
            .entry_by_ref(key)
            .and_compute_with(|maybe_entry| {
                let op = if let Some(entry) = maybe_entry {
                    let counter = entry.into_value();
                    Op::Put(counter.saturating_add(1)) // Update
                } else {
                    Op::Put(1) // Insert
                };
                // Return a Future that is resolved to `op` immediately.
                std::future::ready(op)
            })
            .await
    }

    let cache: Arc<Cache<String, u64>> = Arc::new(Cache::new(100));
    let key = "key1".to_string();

    let mut handles = Vec::new();

    for _ in 0..CONCURRENCY {
        let cache_clone = Arc::clone(&cache);
        let key_clone = key.clone();
        let handle = tokio::spawn(async move {
            for _ in 0..(N / CONCURRENCY) {
                let res = inclement_counter(&cache_clone, &key_clone).await;
                println!("{:?}", res);
            }
        });
        handles.push(handle);
    }

    for handle in handles {
        handle.await.unwrap();
    }

    let result = cache.get(&key).await;
    println!("{result:?}");
}

async fn computation_maybe_cancelled() {
    // Increment a cached `u64` counter.
    async fn inclement_counter_if_nobody_else(
        cache: &Cache<String, u64>,
        key: &str,
    ) -> Result<CompResult<String, u64>, ()> {
        cache
            .entry_by_ref(key)
            .and_try_compute_if_nobody_else(|maybe_entry| {
                let op = if let Some(entry) = maybe_entry {
                    let counter = entry.into_value();
                    Ok(Op::Put(counter.saturating_add(1))) // Update
                } else {
                    Ok(Op::Put(1)) // Insert
                };
                // Return a Future that is resolved to `op` immediately.
                std::future::ready(op)
            })
            .await
    }

    let cache: Arc<Cache<String, u64>> = Arc::new(Cache::new(100));
    let key = "key1".to_string();

    let mut handles = Vec::new();

    for _ in 0..CONCURRENCY {
        let cache_clone = Arc::clone(&cache);
        let key_clone = key.clone();
        let handle = tokio::spawn(async move {
            for _ in 0..(N / CONCURRENCY) {
                let res = inclement_counter_if_nobody_else(&cache_clone, &key_clone)
                    .await
                    .unwrap(); // unwrap safe
                println!("{:?}", res);
            }
        });
        handles.push(handle);
    }

    for handle in handles {
        handle.await.unwrap();
    }

    let result = cache.get(&key).await;
    println!("{result:?}");
}
```

tatsuya6502 (approved):
Thank you very much for contributing this! It looks good to me. (Sorry for taking so long to review)

xuehaonan27:
Thank you for the kind words and for taking the time to review my PR! I completely understand that things can get busy, and I appreciate the effort you put into maintaining the project. I'm glad to see my contribution merged, and I look forward to contributing more in the future if the opportunity arises.

PR #466, opened 2024-12-02
Enable atomic64 only when target supports AtomicU64

### LINKED ISSUE #465: Automatically disable atomic64 feature on known 32-bits Platforms that fail to compile
`moka` depends on `std::atomic::AtomicU64` by feature `atomic64`, which is in the `default` feature list.

If a Project that depends `moka` indirectly, it may not have a chance to set `default-features = false` as suggested in [README](https://github.com/moka-rs/moka#compile-errors-on-some-32-bit-platforms).

I have a proposal:

We can use `#[cfg(target_has_atomic)]` to ensure `std::atomic::AtomicU64` is available in code to make the `atomic64` feature work, just like this:

```rust
#[cfg_attr(
    all(feature = "atomic64", target_has_atomic = "64", feature = "quanta"),
    path = "concurrent/atomic_time/atomic_time.rs"
)]
```

zonyitoo:
- target_has_atomic was stablized in v1.60.0, https://doc.rust-lang.org/beta/releases.html#version-1600-2022-04-07

fix #465

### DIFF
--- a/src/common/concurrent.rs
+++ b/src/common/concurrent.rs
@@ -12,12 +12,8 @@ pub(crate) mod entry_info;
 #[cfg(feature = "sync")]
 pub(crate) mod housekeeper;
 
-// target_has_atomic is more convenient but yet unstable (Rust 1.55)
-// https://github.com/rust-lang/rust/issues/32976
-// #[cfg_attr(target_has_atomic = "64", path = "common/time_atomic64.rs")]
-
 #[cfg_attr(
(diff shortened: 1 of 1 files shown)


zonyitoo:
@tatsuya6502 The 2 failed CI tests are not related to this PR.

zonyitoo:
Are there any suggestions about this PR? Could it be merged?

tatsuya6502 (approved):
Thank you very much for your contribution! It looks good to me.

PR #479, opened 2025-01-02
Documentation: Remove leftover mentions of background threads

### LINKED ISSUE #464: Documentation: Leftover mentions of background threads
The `invalidate_all()` and `invalidate_entries_if()` methods on the sync cache say that a background thread will evict entries. My understanding is that v0.12.0 removed background threads, so this is instead done when handling maintenance operations in other methods.

tatsuya6502:
Fixes #464.

- Remove leftover mentions of background threads from `invalidate_all` and `invalidate_entries_if` methods of all cache implementations.
- Add the following chapters to the crate top level doc page:
    - Cache Policies
    - Implementation Details, with explanations about the channels and maintenance tasks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,12 @@ Bumped the minimum supported Rust version (MSRV) to 1.70 (June 1, 2023)
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


tatsuya6502 (commented):
Merging.

Issue #41, opened 2021-09-14
Remove `atomic64` feature and use `crossbeam::atomic::AtomicCell<u64>` for all targets?


tatsuya6502:
In order to support target platforms without `AtomicU64`, we added a Cargo feature called `atomic64` to Moka v0.5.3 and put it into the default feature list (#38, #39).

- If the feature is enabled, Moka keeps the old behavior (by using `AtomicU64` as timestamps of cache entries).
- Otherwise, it uses `RwLock<Instant>` so that it will compile for such targets.

The feature also controls whether quanta crate is used or not by Moka, because quanta v0.9.2 or earlier do not compile for such targets.

We might want to replace above with `crossbeam::atomic::AtomicCell<u64>` because it provides a similar functionality to `AtomicU64` (when it is supported by the target) and `RwLock<u64>` (when `AtomicU64` is not supported).

https://docs.rs/crossbeam/0.8.1/crossbeam/atomic/struct.AtomicCell.html
> Operations on `AtomicCells` use atomic instructions whenever possible, and synchronize using global locks otherwise.

It will be more convenient as it switches implementations without having a Cargo feature like `atomic64`. It combines auto-generated Rust source and the build script to achieve auto-switching.

Also, the next version of quanta will support these targets (https://github.com/metrics-rs/quanta/pull/55/), so once it is released, we can always enable quanta.

tatsuya6502:
Before switching to `AtomicCell<T>`, we need to check its implementation to see if it as efficient as our current implementation.

tatsuya6502:
> Also, the next version of quanta will support these targets ([metrics-rs/quanta#55](https://github.com/metrics-rs/quanta/pull/55)), so once it is released, we can always enable quanta.

[quanta v0.9.3](https://crates.io/crates/quanta/0.9.3) has been released and Moka v0.5.4 (next version) and newer will always enable quanta (#42).

tatsuya6502:
Started to use `AtomicCell<u64>` for wight based eviction: https://github.com/moka-rs/moka/pull/24/commits/4404e6fd1cca5d41aa59688b7b67c842425c7be3

PR #480, opened 2025-01-04
Add `portable-atomic` crate for the platforms where `AtomicU64` is not available

### LINKED ISSUE #462: unable to build for some platforms that does not support AtomicU64
because moka internally uses some 64-bit atomic types from standard library, but it is not found in some 32-bit processors.
you can use `portable-atomic` for support these platforms.

https://docs.rs/portable-atomic

`cargo add moka --no-default-features -F future,logging,atomic64,quanta`

`cross build --release --target mipsel-unknown-linux-musl`
```
error[E0432]: unresolved import "std::sync::atomic::AtomicU64"
 --> /cargo/registry/src/index.crates.io-6f17d22bba15001f/moka-0.12.8/src/common/concurrent/atomic_time/atomic_time.rs:5:20
  |
5 |     sync::atomic::{AtomicU64, Ordering},
  |                    ^^^^^^^^^
  |                    |
  |                    no "AtomicU64" in "sync::atomic"
  |                    help: a similar name exists in the module: "AtomicU32"
```

tatsuya6502:
Fixes #462, #41.

- Add `portable-atomic` crate for the platforms where `AtomicU64` is not available.
- The crate feature `atomic64` no longer has any effect but we will keep it for now for backward compatibility.
    - It will be removed in `v0.13.0` release.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -101,7 +101,7 @@ jobs:
         run: cargo test --release --lib --features sync sync::cache::tests::ensure_gc_runs_when_dropping_cache -- --exact --ignored
 
       - name: Run tests (future feature, but no sync feature)
-        run: cargo test --no-default-features --features 'future, atomic64, quanta'
+        run: cargo test --features future
 
       - name: Run tests (future, sync and logging features)
         run: cargo test --features 'future, sync, logging'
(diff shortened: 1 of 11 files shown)


tatsuya6502 (commented):
Merging.

Issue #472, opened 2024-12-20
`Option::unwrap()` panics in `base_cache::Clocks::to_std_instant`


matheus23:
In my app I'm using `hickory_resolver`, which uses a DNS resolver, which has a cache that's implemented using `moka`.

I don't think it's using the API incorrectly, but `moka` ends up panicking sometimes. The backtrace for that is:
```
thread 'tokio-runtime-worker' panicked at /home/philipp/.cargo/registry/src/index.crates.io-6f17d22bba15001f/moka-0.12.8/src/sync_base/base_cache.rs:899:59:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/std/src/panicking.rs:662:5
   1: core::panicking::panic_fmt
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:74:14
   2: core::panicking::panic
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:148:5
   3: core::option::unwrap_failed
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/option.rs:2015:5
   4: core::option::Option<T>::unwrap
             at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/option.rs:965:21
   5: moka::sync_base::base_cache::Clocks::to_std_instant
             at /home/philipp/.cargo/registry/src/index.crates.io-6f17d22bba15001f/moka-0.12.8/src/sync_base/base_cache.rs:899:22
   6: moka::sync_base::base_cache::BaseCache<K,V,S>::do_get_with_hash
             at /home/philipp/.cargo/registry/src/index.crates.io-6f17d22bba15001f/moka-0.12.8/src/sync_base/base_cache.rs:327:26
   7: moka::sync_base::base_cache::BaseCache<K,V,S>::get_with_hash
             at /home/philipp/.cargo/registry/src/index.crates.io-6f17d22bba15001f/moka-0.12.8/src/sync_base/base_cache.rs:226:9
   8: moka::sync::cache::Cache<K,V,S>::get
             at /home/philipp/.cargo/registry/src/index.crates.io-6f17d22bba15001f/moka-0.12.8/src/sync/cache.rs:803:9
[...]

PR #481, opened 2025-01-05
Replace most uses of `quanta::Instant` with `std::time::Instant` to increase the accuracy of time measurements

tatsuya6502:
This PR does the following:

- Replace most uses of `quanta::Instant` with `std::time::Instant` to increase the accuracy of time measurements.
    - When the `quanta` feature of `moka` is disabled, `std::time::Instant` is used for all time measurements.
- Refactor time-related code to reduce memory overhead when using `std::time::Instant`.
    - Now the memory overhead of using `std::time::Instant` is the same as using `quanta::Instant`.
    - `moka` keeps three timestamps for each cache entry. (last accessed time, last modified time, and expiration time)
    - Now each timestamp is stored as `AtomicU64` (8 bytes) in nanoseconds.
    - Before, each timestamp was stored as `Mutex<std::time::Instant>`, which is 24 bytes on Rust 1.83 for Linux x86_64 target.
- To avoid unexpected panics on some platforms, use saturated arithmetic for time calculations whenever possible.

`quanta` is a high-performance time measurement library and will be great for some use cases such as collecting performance metrics. However, it is not very suitable for tracking expiration times of cache entries because the time measurement may drift over time, and it may not be monotonic on some specific hardware (#472). Therefore, using `std::time::Instant` will be more appropriate for tracking expiration times of cache entries.

(I created PR #482 to disable the `quanta` feature of `moka` by default)


tatsuya6502:
I ran short benchmarks using mokabench on `moka@v0.12.9` and one on `disable-quanta-by-default` branch. I ran them on my home computers; Linux x86_64 (WSL2) and macOS arm64. I do not see any meaningful performance difference between `moka@v0.12.9` and one on `disable-quanta-by-default` branch. (I only used the default features of `moka`, so `quanta` was disabled on the benchmark using `disable-quanta-by-default` branch)

mokabench-linux-x86_64

mokabench-macos-arm64

tatsuya6502 (commented):
Merging.

PR #482, opened 2025-01-05
Disable the `quanta` feature by default

tatsuya6502:
This PR _disables_ the `quanta` feature by default.

If it is manually enabled, `quanta::Instant` will be used for some performance critical operations in the cache instead of `std::time::Instant`.

The `quanta` crate is a high performance time measurement library and will be great for some use cases such as collecting performance metrics. `moka` used to use `quanta::Instant` all over the place by default. However, `quanta` is not very good for tracking cache entry expiration times because the time measurement can drift over time and it may not be monotonic on some specific hardware (#472). Therefore, using `std::time::Instant` is more appropriate for tracking enty expiration times.

As of `v0.12.10`, the `quanta` feature will not make any noticeable performance difference, but in the future when cache metrics are added, it will be useful to have this feature enabled.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -103,5 +103,8 @@ jobs:
       - name: Run tests (future feature, but no sync feature)
         run: cargo test --features future
 
+      - name: Run tests (release, future feature, but no sync feature)
+        run: cargo test --release --features future
+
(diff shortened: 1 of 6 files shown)


tatsuya6502 (commented):
Merging.

PR #486, opened 2025-01-12
Put the troubleshooting section back in the README

tatsuya6502:
- Put the troubleshooting section back in the README, with some rewritten content:
    - When I released `v0.12.10`, I removed the troubleshooting section because the problems listed in that section should not occur in `v0.12.10` or later.
    - However, I realized that users with older versions may still see the README in this repository, so it is worth keeping the section.
- Remove the FOSSA badge from the README.
    - The FOSSA license scan no longer works for some reason.
- Update the copyright year in the LICENSE files.

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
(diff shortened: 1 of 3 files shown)


tatsuya6502 (commented):
Merging.

Issue #487, opened 2025-01-12
CI: `armv7-unknown-linux-musleabihf` `cross` test fails with Rust 1.84.0


tatsuya6502:
The Linux cross-compiling test for the `armv7-unknown-linux-musleabihf` target has been failing since the release of Rust 1.84.0. It consistently fails on a specific test: `future::base_cache::tests::test_per_entry_expiration`. The same test for the `sync` cache succeeds.

https://github.com/moka-rs/moka/actions/runs/12717383830/job/35456822039

```console
$ cross test --release -F future \
    --target armv7-unknown-linux-musleabihf

...

failures:

---- future::base_cache::tests::test_per_entry_expiration stdout ----
thread 'future::base_cache::tests::test_per_entry_expiration' panicked at src/future/base_cache.rs:3031:25:
assertion `left == right` failed: Mismatched `last_modified_at`s. line: 3176
  left: Instant { tv_sec: 248, tv_nsec: 1673538076 }
 right: Instant { tv_sec: 247, tv_nsec: 388570781 }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    future::base_cache::tests::test_per_entry_expiration

test result: FAILED. 91 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out; finished in 24.87s
```

tatsuya6502:
I can reproduce the failure on my Mac with `cross`.

- It fails when it is cross-compiled by `rustc` 1.84.0 (9fc6b4312 2025-01-07).
- It passes whet it is cross-compiled by `rustc` 1.83.0 (90b35a623 2024-11-26).

Environment:

- `rustc` host: `x86_64-unknown-linux-gnu`
- `rustc` target: `armv7-unknown-linux-musleabihf`.
- macOS Sequoia 15.2 (arm64)
    - Docker: OrbStack 1.9.2 (18814)
- `cross` 0.2.5 (4090bec 2024-10-15)

tatsuya6502:
This seems to be an incorrect compilation by Rust 1.84.0 for the `armv7-unknown-linux-musleabihf` target.

- If I remove the `--release` option, the test passes.
- If I add a `dbg!` print at the following location, the test passes even if the `--release` option is specified.

```diff
diff --git a/src/common/time/clock.rs b/src/common/time/clock.rs
index 2032f91..c63a3eb 100644
--- a/src/common/time/clock.rs
+++ b/src/common/time/clock.rs
[...]

PR #488, opened 2025-01-12
Workaround for incorrect compilation of test code when using Rust 1.84.0 for Linux ARMv7 target

tatsuya6502:
Ref #487


tatsuya6502 (commented):
Merging.

PR #489, opened 2025-01-12
Add the `jittered_expiry_policy` example

### LINKED ISSUE #453: Support TTL and TTI jitter
Hi, I absolutely love the project. We use it in many places.

Would it be possible to allow jitter on the durations? We use moka for various RPC caching. When many entries expire at the same time, we end up sending a big spike to the upstream. I want to jitter ttl duration to combat this issue. I'd love to contribute too, but probably gonna need a few pointers.

tatsuya6502:
- Add the `jittered_expiry_policy_sync` example.
- Add the `rand` crate to the development dependencies.

Fixes #453

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,7 @@ env_logger = "0.10.0"
 getrandom = "0.2"
 once_cell = "1.7"
 paste = "1.0.9"
+rand = "0.8.5"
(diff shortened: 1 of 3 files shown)


tatsuya6502 (commented):
Merging.

Issue #166, opened 2022-07-09
Implications of `Arc<K>: Borrow<Q>`


tatsuya6502:
### Discussed in https://github.com/moka-rs/moka/discussions/163

<div type='discussions-op-text'>

<sup>Originally posted by **JayKickliter** July  9, 2022</sup>
For my particular use case I'm finding `moka::sync::Cache` requirement of `Arc<K>: Borrow<Q>` very restrictive, as my keytype is somewhat expensive to create. Is this an absolute requirement for this kind of cache or an implementation detail?

## Expected behavior

Using `HashMap` as an example, I'd expect to be able to querey using a `&[u8]` when the key type is `Vec<u8>`.

```rust
    use std::collections::HashMap;
    let mut map: HashMap<Vec<u8>, ()> = HashMap::new();
    map.insert(vec![1_u8], ());
    map.contains_key([1_u8].as_slice());
```

## Failing example, simplified

```rust
    use moka::sync::Cache;
    let cache: Cache<Vec<u8>, ()> = Cache::new(1);
    cache.insert(vec![1_u8], ());
    cache.contains_key([1_u8].as_slice());
```

```
error[E0277]: the trait bound `Arc<Vec<u8>>: Borrow<[u8]>` is not satisfied
   --> src/main.rs:5:11
    |
5   |     cache.contains_key([1_u8].as_slice());
    |           ^^^^^^^^^^^^ the trait `Borrow<[u8]>` is not implemented for `Arc<Vec<u8>>`
    |
    = help: the trait `Borrow<T>` is implemented for `Arc<T>`
note: required by a bound in `moka::sync::Cache::<K, V, S>::contains_key`
   --> /Users/jay/.cargo/registry/src/github.com-1ecc6299db9ec823/moka-0.9.0/src/sync/cache.rs:880:17
    |
880 |         Arc<K>: Borrow<Q>,
    |                 ^^^^^^^^^ required by this bound in `moka::sync::Cache::<K, V, S>::contains_key`
```

**EDIT**: I now wonder if this could be considered a bug after noticing the following.

This also seems to imply that the docs [`contains_key`]  are incorrect:

> The key may be any borrowed form of the cache’s key type,

[`contains_key`]: https://docs.rs/moka/latest/moka/sync/struct.Cache.html#method.contains_key</div>

tatsuya6502:
I found a solution for `sync` and `future` caches. I am working on it here:

- #167
[...]

PR #492, opened 2025-01-16
Support `Equivalent` trait for the cache key type `K`

tatsuya6502:
## Summary

Support the `Equivalent` trait from the [equivalent crate](https://crates.io/crates/equivalent) for the cache key type `K`. Replace all usages of the trait bound `K: Borrow<Q>` with `Q: Equivalent<K>`.

For example,

**Before**

```rust
pub fn get<Q>(&self, key: &Q) -> Option<V>
where
    K: Borrow<Q>,
    Q: Hash + Eq + ?Sized,
```

**After**

```rust
pub fn get<Q>(&self, key: &Q) -> Option<V>
where
    Q: Equivalent<K> + Hash + ?Sized,
```

https://github.com/indexmap-rs/equivalent/blob/master/README.md

> These traits are not used by the maps in the standard library, but they may add more flexibility in third-party map implementations, especially in situations where a strict `K: Borrow<Q>` relationship is not available.

## Changes

- Add the [equivalent crate](https://crates.io/crates/equivalent) to the dependencies.
- Reexport the `equivalent` module.
- Replace all trait bound `K: Borrow<Q>` with `Q: Equivalent<K>`.

Relates to #166

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -236,6 +236,9 @@ compile_error!(
     `moka` crate. Please update your dependencies in Cargo.toml"
 );
 
+// Reexport(s)
+pub use equivalent;
+
 #[cfg(feature = "future")]
 #[cfg_attr(docsrs, doc(cfg(feature = "future")))]
 pub mod future;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -237,7 +237,7 @@ compile_error!(
 );
 
 // Reexport(s)
-pub use equivalent;
+pub use equivalent::Equivalent;
 
 #[cfg(feature = "future")]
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -236,6 +236,9 @@ compile_error!(
     `moka` crate. Please update your dependencies in Cargo.toml"
 );
 
+// Reexport(s)
+pub use equivalent;
```
tatsuya6502, line 240: Perhaps, we should reexport just a trait rather than the entire module (crate)?

```suggestion
pub use equivalent::Equivalent;
```

tatsuya6502 (commented):
Merging.

Issue #501, opened 2025-02-28
Switch `loom` to dev-dependency


Blizzara:
Hey, thanks for making and maintaing this crate! I was looking at the dependencies it introduces downstream in my project, and among them is loom:

```
[[package]]
name = "moka"
version = "0.12.10"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "a9321642ca94a4282428e6ea4af8cc2ca4eac48ac7a6a4ea8f33f76d0ce70926"
dependencies = [
 "async-lock",
 "crossbeam-channel",
 "crossbeam-epoch",
 "crossbeam-utils",
 "event-listener",
 "futures-util",
 "loom",
 "parking_lot",
 "portable-atomic",
 "rustc_version",
 "smallvec",
 "tagptr",
 "thiserror 1.0.69",
 "uuid",
]
```

However it seems that `loom` is only used for testing, could it be made a dev-dependency instead? Currently it's a normal dep: https://github.com/moka-rs/moka/blob/e970c8404b2dfa0ad105f8a68e4bcc7815cdad5c/Cargo.toml#L88 but the lines below use dev-dependency.

I may be missing something here in case that isn't possible, but if it is it could be a nice improvement :)

tatsuya6502:
Thank you for reporting this issue. `loom` is changed to a dev-dependency via #509. 

Note: We may change `loom` back to a not-dev, but optional (`optional = true`) dependency in the future. https://github.com/moka-rs/moka/pull/509#issuecomment-2863262532

PR #509, opened 2025-04-20
Switch loom to a dev-dependency

thomaseizinger:
The `loom` dependency is only used for testing and should therefore be a `dev-dependency`. At present, this pulls in an old version `windows` for us which I'd like to get rid of in our dependency tree:

```
❯ cargo tree -i -p windows@0.58.0 --target all
windows v0.58.0
└── generator v0.8.4
    └── loom v0.7.2
        └── moka v0.12.10
            └── firezone-gateway v1.4.7 (/home/thomas/src/github.com/firezone/firezone/rust/gateway)
```

Resolves: #501

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,7 +85,7 @@ tokio = { version = "1.19", features = ["fs", "io-util", "macros", "rt-multi-thr
 
 # We cannot use `cfg(loom)` here because an indirect dependency `concurrent-queue`
 # uses it.
-[target.'cfg(moka_loom)'.dependencies]
+[target.'cfg(moka_loom)'.dev-dependencies]
 loom = "0.7"
 
 [target.'cfg(trybuild)'.dev-dependencies]


thomaseizinger:
@tatsuya6502 Could you look at this please? :pray:

tatsuya6502:
@thomaseizinger — Thank you for submitting this pull request! I will try to review it in a couple of days.

tatsuya6502:
Memo:

When a downstream crate want to run `loom` tests, it will be ideal that its all dependencies (including `moka`) use `loom`'s lock types instead of ones from `std` or `tokio`, etc. This cannot be done if these dependencies have `loom` under dev-dependencies.

I found that `crossbeam-epoch` and `asyc-lock` crates achieve it by combining a crate feature and a conditional (`cfg(...)`), optional (`optional = true`), not-dev dependency to `loom`.

https://github.com/crossbeam-rs/crossbeam/blob/f92b4a71e839c748e944c051a73d387a676fd788/crossbeam-epoch/
Cargo.toml#L30-L44

```toml
[features]
# ...
loom = ["loom-crate", "crossbeam-utils/loom"]

[target.'cfg(crossbeam_loom)'.dependencies]
loom-crate = { package = "loom", version = "0.7.1", optional = true }
```

https://github.com/smol-rs/async-lock/blob/cc249b2d31590a6e97616a351032747f3e19bc87/Cargo.toml#L22-L28

However, I will accept this pull request to make `loom` as a dev-dependency, because currently `moka` uses `loom`'s lock types only in the `arc` module ([code](https://github.com/moka-rs/moka/blob/e970c8404b2dfa0ad105f8a68e4bcc7815cdad5c/src/common/concurrent/arc.rs#L21-L25)), but other lock types in other modules.

tatsuya6502 (approved):
Thank you for submitting this pull request! It looks good to me.

thomaseizinger:
That is very interesting, thank you for the thorough analysis!

Issue #511, opened 2025-05-08
Drop `thiserror` as a dependency


tatsuya6502:
Originally posted by @brownjohnf in https://github.com/moka-rs/moka/issues/506#issuecomment-2741481220

> I was just working on optimizing a dependency tree in a project, and noticed that we were pulling in 2 versions of thiserror, the older one coming from moka.
>
> As I mentioned in one of the commits, I also noticed that we're barely using thiserror here; if you were interested I'd be happy to PR a change to just drop it as a dependency and manually impl Error for the one place (that I've seen) we're using it outside of tests. Maybe we could just move it to a dev-dependency or something.

brownjohnf:
I'm going to propose a PR for this.

brownjohnf:
Opened https://github.com/moka-rs/moka/pull/512

PR #506, opened 2025-03-20
Upgrade thiserror to 2.0 and disable default features

brownjohnf:
Disable default-features on thiserror

We're barely using thiserror at all in moka, and we don't require any of
thiserror's features.

We might also want to consider not using thiserror at all and pruning
that whole dependency tree.

Upgrade thiserror to 2.0

Thiserror has had a 2.0 release that does include breaking changes but
does not appear to impact moka's use of thiserror.

By upgrading, we can stay closer to the latest version of thiserror and
avoid needing to compile multiple versions of thiserror in consumers who
are themselves using a newer version of thiserror.

PR #512, opened 2025-05-08
Replace thiserror usage with manual impl

brownjohnf:
We were only using thiserror in 2 places, and one of them was in tests. Our usage was also very minimal (no use of the #[from] directive, etc.), so I think it makes more sense to just manually implement Error for these limited uses.

It's hard to see the impact here because we also don't check-in Cargo.lock, but locally running a cargo tree, I see 5 fewer lines of dependencies.

I went down this path because while auditing an unrelated project I maintain, I noticed that we were pulling in 2 copies of thiserror because moka required version 1, but the other project was pulling in version 2. I opened #506 to address that, but that doesn't help any projects out there who are using version 1. Sometimes crates add dedicated version feature flags like thiserror_v1 and thiserror_v2 or something, but given our minimal usage I thought we could just remove it completely.

Closes https://github.com/moka-rs/moka/issues/511

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,6 @@ parking_lot = "0.12"
 portable-atomic = "1.6"
 smallvec = "1.8"
 tagptr = "0.2"
-thiserror = "1.0"
 uuid = { version = "1.1", features = ["v4"] }
 
 # Optional dependencies (quanta)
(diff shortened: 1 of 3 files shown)


tatsuya6502 (approved):
Thanks for working on this! It looks good to me.

brownjohnf:
Glad to help!

PR #516, opened 2025-05-15
Allow key type to be unsized and prefer str to String as key type

kxxt:
**Rationale**

Currently, moka wraps the key type inside an `Arc`. Coupled with the usage of `ToOwned` trait,
this creates a two level pointer indirection and waste of some space if we use `String` as the key type, which is commonly used.
This PR creates a new trait `ToOwnedArc` which converts the type to be optimally wrapped inside an `Arc`
and then removes the `Sized` requirement on key type.

**Warning**

Although in general, `Arc<String>` seems to be an anti-pattern. But for some workload that already has a `String`,
it could avoid the potential reallocation when converting that `String` into `Arc<str>`.
So this PR might bring some negative performance effects.

And this PR of course introduces BREAKING changes.

**Benchmark**

Unfortunately, mokabench does not have benches for string key type. So I currently don't have any benchmark results to back this PR.
Running mokabench with this PR produces similar results without this PR, though.


tatsuya6502:
Hi. Thank for clarifying. Now I understand the breaking changes better. OK. I will release it as `v0.13.0` instead of `v0.12.x`.

As for the new generic parameter `OPTIMAL`, I am not sure what would be the best, but I came up with two additional options:

- **Other option 1**: Do not expose it to the user.
    - Instead, create a thin wrapper at the top level of the `Cache` and do the selection internally.
    - Maybe for the `Cache` struct, it will always be `false`, so the behavior is the same as before.
    - And then create a new struct, e.g., `UnsizedKeyCache`, which always uses `true` and work only for unsized keys such as `str`.
- **Other option 2**: Use a crate like [`castaway`](https://crates.io/crates/castaway) to remove the `OPTIMAL` parameter.
    - `castaway` provides downcasting at compile time.
    - We can use single trait method implementation for any type `T`, and inside the trait method, it can select specialized implementation based on whether `T` is `String` or not.
    - (I have not tried it, so I am not sure if it works)

PR #531, opened 2025-09-18
Update reqwest dev-dependency from 0.11 to 0.12

musicinmybrain:
This allows the tests to be built with the latest version of `reqwest`. I confirmed that `cargo test -F sync,future` still passes.


tatsuya6502 (approved):
Thank you!

Issue #527, opened 2025-08-22
Panic in timer wheel


conradludgate:
I haven't had a chance to investigate the cause of the failure yet, but I've just observed this panic:

```
internal error: entered unreachable code

Thread 0 Crashed:
0   std                             0x5649bf4be649      std::sys::backtrace::__rust_end_short_backtrace (backtrace.rs:168)
1   __rustc                         0x5649bf4bff1d      __rustc::rust_begin_unwind (panicking.rs:697)
2   core                            0x5649bf4e7430      core::panicking::panic_fmt (panicking.rs:75)
3   core                            0x5649bf4e74bc      core::panicking::panic (panicking.rs:145)
4   moka                            0x5649be331b23      moka::common::timer_wheel::TimerWheel<T>::unlink_timer (timer_wheel.rs:283)
5   moka                            0x5649be3318a2      moka::common::timer_wheel::TimerWheel<T>::deschedule (timer_wheel.rs:265)
6   moka                            0x5649be2a67d7      [inlined] moka::sync_base::base_cache::Inner<T>::handle_upsert (base_cache.rs:1473)
7   moka                            0x5649be2a67d7      [inlined] moka::sync_base::base_cache::Inner<T>::apply_writes (base_cache.rs:1423)
8   moka                            0x5649be2a67d7      moka::sync_base::base_cache::Inner<T>::do_run_pending_tasks (base_cache.rs:1212)
9   moka                            0x5649be3b075e      [inlined] moka::sync_base::base_cache::Inner<T>::run_pending_tasks (base_cache.rs:1163)
10  moka                            0x5649be3b075e      [inlined] moka::common::concurrent::housekeeper::Housekeeper::do_run_pending_tasks (housekeeper.rs:125)
11  moka                            0x5649be3b075e      moka::common::concurrent::housekeeper::Housekeeper::try_run_pending_tasks (housekeeper.rs:112)
12  moka                            0x5649be2b9969      [inlined] moka::sync_base::base_cache::BaseCache<T>::apply_reads_if_needed (base_cache.rs:611)
13  moka                            0x5649be2b9969      [inlined] moka::sync_base::base_cache::BaseCache<T>::record_read_op (base_cache.rs:479)
14  moka                            0x5649be2b9969      moka::sync_base::base_cache::BaseCache<T>::get_with_hash::{{closure}} (base_cache.rs:225)
[...]

Issue #524, opened 2025-06-27
segfault in expiry/timerwheel


sharksforarms:
Hi! We noticed a segfault from moka in our application. Here's a backtrace:

```
   1: moka::common::concurrent::entry_info::EntryInfo<K>::expiration_time
             at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/moka-0.12.10/src/common/concurrent/entry_info.rs:121:9
      moka::common::timer_wheel::TimerWheel<K>::schedule_existing_node
             at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/moka-0.12.10/src/common/timer_wheel.rs:235:49
   2: moka::common::timer_wheel::TimerWheel<K>::reschedule
             at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/moka-0.12.10/src/common/timer_wheel.rs:258:9
      moka::sync_base::base_cache::Inner<K,V,S>::update_timer_wheel
             at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/moka-0.12.10/src/sync_base/base_cache.rs:1753:42
   3: moka::sync_base::base_cache::Inner<K,V,S>::apply_reads
             at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/moka-0.12.10/src/sync_base/base_cache.rs:1392:25
      moka::sync_base::base_cache::Inner<K,V,S>::do_run_pending_tasks
             at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/moka-0.12.10/src/sync_base/base_cache.rs:1207:21
   4: <moka::sync_base::base_cache::Inner<K,V,S> as moka::common::concurrent::housekeeper::InnerSync>::run_pending_tasks
             at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/moka-0.12.10/src/sync_base/base_cache.rs:1163:9
      moka::common::concurrent::housekeeper::Housekeeper::do_run_pending_tasks
             at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/moka-0.12.10/src/common/concurrent/housekeeper.rs:125:29
      moka::common::concurrent::housekeeper::Housekeeper::run_pending_tasks
             at /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/moka-0.12.10/src/common/concurrent/housekeeper.rs:107:9
[...]

PR #548, opened 2025-12-05
Fix use-after-free panic in timer wheel when Expiry returns None

awarus:
Problem:
A race condition exists when using custom Expiry implementations that return None to unset per-entry expiration. expiration_time in EntryInfo is modified atomically when Expiry returns None, but timer_node in DeqNodes is only updated during housekeeping. This creates a window where the expiration time indicates "no expiration" but the timer node pointer still references freed memory.

Solution:
Add a generation counter expiry_gen to EntryInfo.
Counter incremented each time expiration changes
and skip operations on stale nodes instead of panicking (best effort). Long term solution would be apply the set_expiration_time() changes during housekeeping or use arena-based allocation, but it changes external behavior and I consider this as bigger architectural change.

This PR also adds a two tests to catch a such panic and reproduce the issue.

Fixes https://github.com/moka-rs/moka/issues/527 and https://github.com/moka-rs/moka/issues/524.

### DIFF
--- a/src/common/concurrent.rs
+++ b/src/common/concurrent.rs
@@ -140,6 +140,9 @@ pub(crate) struct DeqNodes<K> {
     access_order_q_node: Option<KeyDeqNodeAo<K>>,
     write_order_q_node: Option<KeyDeqNodeWo<K>>,
     timer_node: Option<DeqNodeTimer<K>>,
+    /// The expiry generation when timer_node was set.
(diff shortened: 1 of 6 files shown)


tatsuya6502:
Just for sure, I'm going to ask GitHub Copilot for code review. If the feedback comments make sense, I'll address them in other pull requests.

PR #552, opened 2025-12-14
CI: Fix the CI for the MSRV 1.70.0

tatsuya6502:
When run the CI jobs with MSRV 1.70.0, downgrade some indirect dependencies to versions compatible with it.


tatsuya6502 (commented):
Merging.

PR #592, opened 2026-04-30
Fix LRU eviction stall under concurrent key reuse

### LINKED ISSUE #590: Unbounded memory growth in `future::Cache` — eviction stalls permanently with key reuse under write pressure
## Summary

Thanks for moka — we've been relying on it in production and it's been great. We ran into an issue where `future::Cache` stops evicting entries and memory grows without bound, eventually leading to OOM. After some investigation, we found it happens when there's sustained write pressure and some keys are reinserted (i.e., the same key is inserted more than once over the lifetime of the cache).

Once eviction stalls, `weighted_size()` keeps climbing and calling `run_pending_tasks()` has no effect — the cache never recovers. We were able to write a minimal reproduction (included below) that reliably triggers the issue within 30 seconds.

Workloads that only use unique keys (no reuse) don't seem to be affected, even under very high write rates.

## Version

- **Crate:** moka 0.12.15
- **Feature:** `future`

## Impact

We found a suspected OOM kill in our production environment and were able to reproduce the same unbounded memory growth pattern in a testing environment. The test below is our effort to reproduce the issue in an isolated, minimal form.

## Triggering Conditions

Two conditions must be present simultaneously:

1. **Key reuse** — some keys are inserted more than once (same key, new value)
2. **Sustained write pressure** — enough concurrent writers to race with housekeeping

Unique-only keys (no reuse) do **not** trigger the stall regardless of write rate.

## Reproduction

Save the following as `tests/eviction_stall.rs` and run:

```bash
cargo test --all-features --test eviction_stall -- --ignored --nocapture
```

```rust
#![cfg(all(test, feature = "future"))]

use moka::{future::Cache, policy::EvictionPolicy};
use std::sync::Arc;
use std::time::{Duration, Instant};

/// Eviction stalls permanently when repeated keys are reinserted under
/// concurrent write pressure.
///
/// Setup:
///   - 10 MiB cache, 64 KiB values (~160 entries at capacity)
///   - 10 writers reinserting from a pool of 500 repeated keys
///   - 10 writers inserting unique keys (monotonically increasing)
///   - 30 seconds of sustained writes, then drain
///
/// Expected: weighted_size() <= max_capacity after run_pending_tasks()
/// Actual:   weighted_size() grows without bound, drain has no effect
#[tokio::test(flavor = "multi_thread", worker_threads = 8)]
#[ignore]
async fn eviction_stall_with_key_reuse() {
    let cache: Cache<u64, Vec<u8>> = Cache::builder()
        .max_capacity(10 * 1024 * 1024) // 10 MiB
        .weigher(|_k, v: &Vec<u8>| v.len() as u32)
        .eviction_policy(EvictionPolicy::lru())
        .build();

    let start = Instant::now();
    let duration = Duration::from_secs(30);
    let writers_per_group = 10;
    let barrier = Arc::new(tokio::sync::Barrier::new(writers_per_group * 2));

    // Writers: repeated keys (cycle through 0..500)
    let mut handles = Vec::new();
    for _ in 0..writers_per_group {
        let cache = cache.clone();
        let barrier = Arc::clone(&barrier);
        handles.push(tokio::spawn(async move {
            barrier.wait().await;
            let mut key = 0u64;
            while start.elapsed() < duration {
                cache.insert(key, vec![0u8; 64 * 1024]).await;
                key = (key + 1) % 500;
            }
        }));
    }

    // Writers: unique keys (never repeated)
    for tid in 0..writers_per_group {
        let cache = cache.clone();
        let barrier = Arc::clone(&barrier);
        let base = (tid as u64 + 1) * 1_000_000_000;
        handles.push(tokio::spawn(async move {
            barrier.wait().await;
            let mut key = base;
            while start.elapsed() < duration {
                cache.insert(key, vec![0u8; 64 * 1024]).await;
                key += 1;
            }
        }));
    }

    // Wait for all writers to finish
    for h in handles {
        let _ = h.await;
    }

    // Drain: repeatedly call run_pending_tasks to let housekeeping catch up
    for _ in 0..64 {
        tokio::time::sleep(Duration::from_millis(100)).await;
    }

    let ws = cache.weighted_size();
    let max = 10 * 1024 * 1024;
    eprintln!(
        "weighted_size = {:.1} MiB (max_capacity = {:.1} MiB)",
        ws as f64 / (1024.0 * 1024.0),
        max as f64 / (1024.0 * 1024.0),
    );

    assert!(
        ws <= max,
        "eviction stalled: weighted_size {:.1} MiB exceeds max_capacity {:.1} MiB \
         after 64 rounds of run_pending_tasks()",
        ws as f64 / (1024.0 * 1024.0),
        max as f64 / (1024.0 * 1024.0),
    );
}
```

## Expected Behavior

After writers stop and `run_pending_tasks()` is called 64 times, `weighted_size()` should converge to `<= max_capacity` (10 MiB).

## Actual Behavior

`weighted_size()` grows to tens of GiB and never decreases:

```
weighted_size = 23361.9 MiB (max_capacity = 10.0 MiB)

thread 'eviction_stall_with_key_reuse' panicked at tests/eviction_stall.rs:
eviction stalled: weighted_size 23361.9 MiB exceeds max_capacity 10.0 MiB
after 64 rounds of run_pending_tasks()
```

## Notes

- **`weighted_size()`** — the cache's own accounting — confirms the stall. This is not just a transient lag.
- **Drain produces zero progress** — `run_pending_tasks()` cannot recover once the stall occurs.
- **Unique-key-only workloads are unaffected** — verified over a 1-hour sustained test with 540 million inserts; `weighted_size()` stays at exactly `max_capacity` throughout.

---

Happy to provide more details, run additional tests, or try patches if that would help. Thanks for looking into this.

kim-jhyeon:
A stale `WriteOp::Upsert` drained for an already-evicted entry pushed an orphan deque node that the LRU eviction loop could not reclaim, stalling eviction indefinitely and letting `weighted_size` grow without bound.

Add an `is_retired` flag to `EntryInfo`, set inside the post-CAS callback of `cht::remove_entry_if_and` at every removal site so the retirement linearises with the bucket unlink. Buffered-op consumers (`handle_upsert`, `apply_reads`'s Hit handler, TinyLFU inline victim scan) check `is_alive()` before mutating policy state. Includes regression tests for both sync and future variants.

Fixes #590.

---

<a href="https://app.devin.ai/review/moka-rs/moka/pull/592" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added entry lifecycle management to prevent stale operations on removed cache entries.

* **Bug Fixes**
  * Fixed concurrent eviction stalls by ensuring cache policies skip retired entries during operation processing.

* **Tests**
  * Added regression tests validating eviction convergence under concurrent key reuse patterns in both sync and async caches.

### DIFF
--- a/src/cht/segment.rs
+++ b/src/cht/segment.rs
@@ -337,6 +337,9 @@ impl<K: Hash + Eq, V, S: BuildHasher> HashMap<K, V, S> {
     /// Removes a key from the map, returning a clone of the key-value pair
     /// previously corresponding to the key.
     #[inline]
+    #[allow(dead_code)] // Kept for the CHT's public-ish API surface; current
+                        // moka call sites go through `remove_entry_if_and` so the
+                        // post-CAS callback can retire the `EntryInfo`.
     pub(crate) fn remove_entry(&self, hash: u64, eq: impl FnMut(&K) -> bool) -> Option<(K, V)>
     where
         K: Clone,
--- a/src/common/concurrent/entry_info.rs
+++ b/src/common/concurrent/entry_info.rs
@@ -10,7 +10,72 @@ pub(crate) struct EntryInfo<K> {
     /// `is_admitted` indicates that the entry has been admitted to the cache. When
     /// `false`, it means the entry is _temporary_ admitted to the cache or evicted
     /// from the cache (so it should not have LRU nodes).
+    ///
+    /// This is distinct from [`is_retired`](Self::is_retired): `is_admitted`
+    /// tracks deque membership, while `is_retired` tracks CHT membership. A
+    /// freshly inserted entry is **not retired** and **not admitted** until its
+    /// `WriteOp::Upsert` is drained; a retired entry may still be `is_admitted`
+    /// until `handle_remove*` unlinks it.
     is_admitted: AtomicBool,
+    /// CHT-membership flag: `false` while the entry is alive (still backed by
+    /// the concurrent hash table), `true` once the entry has been removed.
+    ///
+    /// The flag is set by `retire()` at the moment the entry is unlinked from
+    /// the CHT and is never cleared (`false → true`, monotonic). The
+    /// invariant it enforces is:
+    ///
+    /// > If `is_retired == true`, no policy-tier mutation may run against
+    /// > this `EntryInfo`.
+    ///
+    /// Examples of policy-tier mutations:
+    ///
+    /// - admitting the entry to an LRU deque,
+    /// - moving its deque node to the back,
+    /// - scheduling a timer-wheel node,
+    /// - incrementing `weighted_size`.
+    ///
+    /// ### Why we need it
+    ///
+    /// moka stores entries in a lock-free CHT (source of truth) and mirrors
+    /// their ordering into LRU deques updated asynchronously via a bounded
+    /// `WriteOp` channel. Without this flag, a `WriteOp::Upsert` can sit in the
+    /// channel long enough that the backing CHT entry is evicted before the op
+    /// is drained; the drain then pushes an "orphan" deque node that the
+    /// eviction loop cannot reach, blocking progress and stalling eviction
+    /// indefinitely (see <https://github.com/moka-rs/moka/issues/590>). The
+    /// retire store is performed inside the CHT's post-success
+    /// `with_previous_entry` callback so the transition linearises with the
+    /// bucket-pointer CAS that physically unlinked the entry; any consumer
+    /// draining a stale op thereafter observes `is_retired == true` and
+    /// declines to touch policy state.
+    ///
+    /// ### Why a single bit suffices (no `Dead` state)
+    ///
+    /// Caffeine encodes a three-state lifecycle (`Alive → Retired → Dead`)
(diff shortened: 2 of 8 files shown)


kim-jhyeon:
It looks like the MSRV tests are failing because `indexmap = ^2.13.0` requires Rust 1.82.0, while Moka’s MSRV is 1.71.1.

tatsuya6502 (approved):
Thank you very much for your time and effort! It looks good to me.

kushudai:
Would it be possible to get a release out for this fix? Thank you!

tatsuya6502:
Yes. I'm trying to do so. I'm trying to fix a panic introduced by this PR now via #599.

tatsuya6502:
@kushudai — Published [moka@v0.12.16](https://crates.io/crates/moka/0.12.16) with this fix to crates.io!

PR #595, opened 2026-06-10
Cost aware LFU

rs-sac:
This series of patches contains a relatively simple variation on the existing LFU algorithm that compares a potential admission to existing cache items based on a weighted cost rather than straight frequency of use.  The idea is that if the purpose of the cache is to reduce time spent computing, it may be the case that an item less frequently used is more worth caching if the time required to recompute its value is disproportionately greater.

(Concretely, if item A is used 1x/sec and item B is used 2x/sec, LFU would prefer B.  But if A costs 10 to recompute and B costs 1 to recompute, then preferring A would save more computation.)

Thus, this new algorithm compares weighted costs rather than frequencies, where the weighted cost is the frequency multiplied by a cost function (similar to the existing weighing function).

---

<a href="https://app.devin.ai/review/moka-rs/moka/pull/595" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added a **Cost-aware TinyLFU** eviction policy variant that incorporates per-entry recomputation cost into admission and eviction decisions (frequency × cost).
  * Added a new `cost()` method to the cache builder to supply per-entry recomputation cost for the cost-aware policy.

* **Tests**
  * Added/updated unit tests to validate cost-aware admission/eviction outcomes.

* **Documentation**
  * Updated eviction/admission policy documentation to describe the cost-aware behavior.


ben-manes:
fwiw, there was some [research](https://www.usenix.org/conference/nsdi26/presentation/keren) in this area that might be interesting. [Sometimes](https://github.com/zilliztech/GPTCache/pull/680#issuecomment-4234477540) the simple and obvious approach works, sometimes it doesn't. Most of the time, caches are configured and not monitored for tuning so I am always hesitant to give eviction algorithm knobs that may be worse if set. I think its also worthwhile for a cache to emphasize that users would need to monitor for the new goal (from hit rate to request latency) since its less intuitive about what good performance results are.

